PR #21, opened 2015-03-21
Don't warn if "Emacs" in the package summary is followed by "Lisp"

Fanael:
Making it a PR because I'm not sure if we want it.

To me, it's a good idea, because while "Emacs" is obvious and thus always redundant, "Emacs Lisp" is not. A good example would be [auto-compile](http://melpa.org/#/auto-compile), which has a pretty generic name, but is Emacs Lisp specific, so it's right for the summary to specify it's Emacs Lisp, er, specific.

What's your thoughts?


Fanael:
Ping.

purcell:
Thanks.

PR #29, opened 2016-10-08
Check for non-standard keywords

alphapapa:
As mentioned in https://github.com/melpa/melpa/pull/4304  :)


purcell:
Yay, that's great - thanks!

alphapapa:
I just realized something: if `finder-by-keyword` (`C-h p`) hasn't been run yet, `finder-known-keywords` will be undefined.  I just pushed 9fc87c9 which requires `finder`, which defines that var.  It seems to fix the problem.

purcell:
Added - thanks!

PR #30, opened 2016-10-12
Check symbols for forbidden characters

alphapapa:
I see a lot of MELPA PRs where you guys have to remind people (including me ;) to not use `/` as a separator in symbols.  Maybe this will help.  It compares symbol names against a list of forbidden characters, which can be extended easily by adding to the `rx` form in the `flycheck-package--forbidden-symbol-chars-regexp` const.

This loads `semantic` so it can call `semantic-fetch-tags`, which seemed like the best way to get a list of symbols and their positions in the current buffer, as it's included with Emacs.

Thanks for your work on MELPA.  :)

### DIFF
--- a/flycheck-package.el
+++ b/flycheck-package.el
@@ -42,6 +42,7 @@
 (require 'package)
 (require 'lisp-mnt)
 (require 'finder)
+(require 'semantic/bovine/el)
 
 
 ;;; Compatibility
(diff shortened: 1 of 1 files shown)


Fanael:
NAK: doesn't appear to do anything here.

alphapapa:
Hm, that's strange.  When I add something like:
```el
(defun test/function ()
  "test"
  nil)
```
...to a buffer, it's marked as a warning, and when I put the point on the first character in the marked line, the error is displayed in the minibuffer.

PR #36, opened 2016-10-13
Verify that defs start with package prefix

alphapapa:
This verifies that symbols defined in the package begin with the package's prefix.  The prefix is assumed to be the symbol the package `provide`s, without any `-mode` suffix.


Fanael:
Thanks!

PR #41, opened 2016-11-04
readme : Corrent package name in credits

rdotdk:
The package name is package-lint, so i think the credits section should mention the same name. Probably a copy paste error?


purcell:
Thanks

PR #43, opened 2016-11-09
List the valid keywords from `finder-known-keywords'

dotemacs:
The function for pointing out the invalid keywords used to send the user
to look up the valid keywords from `finder-known-keywords'. But why have
this extra step, why not just tell the user what the valid keywords are?

This is my attempt at doing just that, pointing out the invalid keyword
and then listing the valid keywords that the user can chose from in
order to pass the linting.

### DIFF
--- a/package-lint.el
+++ b/package-lint.el
@@ -158,6 +158,20 @@ Package-Version headers are present."
 
 ;;; Checks
 
+(defun package-lint--list-known-keywords ()
+  "List the known valid keywords for for the Keyword header.
+Takes the keywords from `finder-known-keywords' and formats them for displaying."
+  (let ((good-known-keywords finder-known-keywords)
+	(formatted-valid-keywords ""))
+    (while good-known-keywords
+      (setq formatted-valid-keywords
+	    (concat formatted-valid-keywords
+		    (format "%11s - %s\n"
+			    (caar good-known-keywords)
+			    (cdr (car good-known-keywords)))))
+      (setq good-known-keywords (cdr good-known-keywords)))
+    formatted-valid-keywords))
+
 (defun package-lint--check-keywords-list ()
   "Verify that package keywords are listed in `finder-known-keywords'."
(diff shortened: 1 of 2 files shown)


dotemacs:
I can see that I have conflicts, but before I start resolving them:

Do you like what I've done here?

Thank you

package-lint.el
```diff
@@ -158,6 +158,20 @@ Package-Version headers are present."
...
+	(formatted-valid-keywords ""))
+    (while good-known-keywords
+      (setq formatted-valid-keywords
+	    (concat formatted-valid-keywords
+		    (format "%11s - %s\n"
+			    (caar good-known-keywords)
+			    (cdr (car good-known-keywords)))))
+      (setq good-known-keywords (cdr good-known-keywords)))
+    formatted-valid-keywords))
+
```
Fanael: Why not simply `(mapconcat (lambda (keyword) (format "%11s - %s\n" (car keyword) (cdr keyword))) finder-known-keywords "")`?

Fanael:
NAK from me, a forty line diagnostic message is way too long for my taste.

dotemacs:
Should it be about "taste" or helping people to write better Emacs packages?

Should the message maybe list the keywords without the explanations? Something like: "keyword, keyword,..."?

purcell:
Thanks, but I'm also -1 on this because it's getting too verbose, sorry. But perhaps an improvement to the current message by making it clearer that `finder-known-keywords` is a variable, and/or including an example?

dotemacs:
Sorry, just saw your comment @purcell.

I'll open an issue since this is closed.

Fanael:
> Should it be about "taste" or helping people to write better Emacs packages?

The problem is that a message this verbose would be ignored by just about everyone, so it would fail to achieve the goal of helping people write better packages.

I agree that the current message can be improved, but I doubt this is the right way.

PR #45, opened 2016-11-10
List the valid keywords

dotemacs:
This PR is built on top of this PR discussion: https://github.com/purcell/package-lint/pull/43

It lists the valid keywords when an invalid keyword is detected.

Yes, the PR can't be merged, but I'm looking for comments if the warning message is valid.

Thank you

### DIFF
--- a/package-lint-test.el
+++ b/package-lint-test.el
@@ -54,45 +54,8 @@ FORCE is passed directly to `package-lint-buffer', which see."
 (ert-deftest package-lint-test-warn-nonstandard-keyword ()
   (should
    (equal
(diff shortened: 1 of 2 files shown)


purcell:
I still strongly prefer the current version, sorry. People will just add a likely-looking valid keyword to the list in response to this message, without necessarily knowing what it denotes, and the right thing to do is to refer people off to the relevant documentation, ie. `finder-known-keywords`. I've made this slightly clearer in 69d5637.

dotemacs:
Not a problem, thanks

PR #46, opened 2016-11-11
Lint always, even in absence of Version and Package-Requires

Fanael:
The idea is that there's no point in calling package-lint when linting is not desired in the first place.

Additionally, expose a predicate that checks if a buffer looks like a package so that automated linting tools like flycheck-package can keep their old behavior.

---

@purcell, can you rebase it onto master when purcell/flycheck-package#2 is merged?


purcell:
Aweseome, thanks.

PR #48, opened 2016-11-24
Ignore sane package prefixes

### LINKED ISSUE #47: Ignore `org-dblock-write:` prefix
org mode use the prefix `org-dblock-write:` to define [dynamic blocks](http://orgmode.org/manual/Dynamic-blocks.html). I think could be safely dismissed as error.

Currenly, with flycheck-package:

```
206   1 error           `org-dblock-write:gargantua' contains a non-standard separator `:', use hyphens instead. (emacs-lisp-package)
206   1 error           "org-dblock-write:gargantua" doesn't start with package's prefix "gargantua". (emacs-lisp-package)
```

marsam:
Fixes #47


Fanael:
Thanks!

purcell:
Thanks @marsam & @Fanael !

PR #53, opened 2017-02-20
List one symbol per line and sort alphabetically

xiongtx:
This makes it easy to pick out repeated symbols. It also makes it easier to
tell, at a glance, if any symbols are out of place or missing.


Fanael:
Thanks!

PR #54, opened 2017-02-21
Check for symbols from cl.el

AdamNiederer:
Checks if any symbols from cl.el (extracted from my copy of Emacs
25.1) are present in the file as a quoted symbol or function call, and
instructs the user to use the equivalent function in cl-lib instead.

To Do:
* [ ] Add test cases
* [ ] Possibly compare cl.el symbols across versions with `(mapatoms)`
* [ ] Find a better way to depropertize strings from the buffer (or
  perhaps leave properties in as a hint?)
* [ ] Figure out a way to not detect local variables

### DIFF
--- a/package-lint.el
+++ b/package-lint.el
@@ -214,6 +214,214 @@ This is bound dynamically while the checks run.")
           with-file-modes)))
   "An alist of function/macro names and when they were added to Emacs.")
 
+(defconst package-lint--unprefixed-cl-symbols
+  '(cddar
+    make-random-state
+    do-all-symbols
+    lexical-let
+    define-modify-macro
+    nsubstitute-if
+    equalp
+    tree-equal
+    cdddr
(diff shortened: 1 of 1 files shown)


Fanael:
I don't like it, I'd rather warn on `(require 'cl)`. Besides, doesn't the byte compiler already warn against calling functions from `cl`?

AdamNiederer:
I didn't know about the byte compiler's built-in cl.el warnings. We could use the byte compiler's output and symbol table in the output of package-lint to eliminate the false positives, actually. Would something like that (parsed or otherwise) in the output be a welcome addition?

purcell:
Yup, I feel like this is out-of-scope for `package-lint`: `flycheck` would cover this while writing code, and any reasonable CI build process would check for clean byte compilation as well as `package-lint` warnings.

Issue #56, opened 2017-03-23
Help configuring/fine-tuning package-lint


kaushalmodi:
Hi,

From the Melpa README.md, I trying following:

> * Use [flycheck-package](https://github.com/purcell/flycheck-package)
  to help you identify common errors in your package metadata.

Here is the .el file in question: https://github.com/kaushalmodi/ox-minutes/blob/master/ox-minutes.el

That gives me these errors and warnings:

```
     1   1 error           `ox-minutes.el --- Plain text backend for Org for Meeting Minutes -*- lexical-binding: t; -*-' contains a non-standard separator `:', use hyphens instead. (emacs-lisp-package)
     1   1 error           "ox-minutes.el --- Plain text backend for Org for Meeting Minutes -*- lexical-binding: t; -*-" doesn't start with package's prefix "ox-minutes". (emacs-lisp-package)
     8   1 warning         You should include standard keywords: see the variable `finder-known-keywords'. (emacs-lisp-package)
    24   1 error           `Commentary:' contains a non-standard separator `:', use hyphens instead. (emacs-lisp-package)
    24   1 error           "Commentary:" doesn't start with package's prefix "ox-minutes". (emacs-lisp-package)
    90   1 error           `Code:' contains a non-standard separator `:', use hyphens instead. (emacs-lisp-package)
    90   1 error           "Code:" doesn't start with package's prefix "ox-minutes". (emacs-lisp-package)
   112   1 error           "Filter functions" doesn't start with package's prefix "ox-minutes". (emacs-lisp-package)
   114   1 error           "org-minutes-final-function" doesn't start with package's prefix "ox-minutes". (emacs-lisp-package)
   136   1 error           "End-user functions" doesn't start with package's prefix "ox-minutes". (emacs-lisp-package)
   139   1 error           "org-minutes-export-as-ascii" doesn't start with package's prefix "ox-minutes". (emacs-lisp-package)
   175   1 error           "org-minutes-export-to-ascii" doesn't start with package's prefix "ox-minutes". (emacs-lisp-package)
   210   1 error           "ox-minutes.el ends here" doesn't start with package's prefix "ox-minutes". (emacs-lisp-package)
```

But most of them do not make sense.. An error is saying that the "Commmentary:" part in the comment header `;;; Commentary:`. Also if I remove that colon, `M-x checkdoc` fails.

Running `M-x package-lint-current-buffer` gives:

```
13 issues found:

[...]

PR #57, opened 2017-03-24
Ignore list elements returns by imenu if they are comments

kaushalmodi:
Users can have imenu set up to list comment headers like

    ;;; Section 1
    ;;;; Section 1.1

Before this patch, "Section 1" and "Section 1.1" were considered as
"variable/function definitions" as per `package-lint--get-defs`.

After this patch, such comment lines are not returned by the variable/function
def list returned by that function.

### DIFF
--- a/package-lint.el
+++ b/package-lint.el
@@ -626,7 +626,11 @@ The returned list is of the form (SYMBOL-NAME . POSITION)."
          (when (member submenu-name '("Variables" "Defuns"))
            (setq result (nconc (reverse submenu-elements) result))))
         (_
-         (push entry result))))
+         (unless (and (consp entry)     ;Do not push entry to result if it's a comment
+                      (save-excursion
+                        (goto-char (cdr entry))
+                        (looking-at-p ";+")))
+           (push entry result)))))
     ;; If it's Semantic, then it returns overlays, not positions. Convert
     ;; them.
     (dolist (entry result)


kaushalmodi:
This PR resolves this issue: https://github.com/purcell/package-lint/issues/56

purcell:
While it might work, I don't think this is the right solution, sorry. We're already looking at only the portions of `imenu`'s results that are listed as `"Variables"` or `"Defuns"`.

Now, you're using `outshine`, which unfortunately causes things that are not variables or defuns to appear as part of those results. That's really an issue with `outshine`, and not something for us to work around here on an ad-hoc basis IMO.

Any thoughts, @Fanael ?

kaushalmodi:
> Now, you're using outshine, which unfortunately causes things that are not variables or defuns to appear as part of those results. 

No, the `outshine` added stuff goes to the "default" or `(_` condition in that pcase. So I have added an `unless` in that condition.

purcell:
Sorry, yes, I misread the patch. But I think the general point may still hold: I'm wary of trying to handle in `package-lint` all the possible imenu entries that might not be vars/defuns. The better solution may be to not rely on `imenu`, i.e. use `semantic` unconditionally.

Fanael:
I'd say that outshine should change its imenu pattern to put its entries into a separate category, because section headers listed as functions look just plain bad in `helm-imenu`.

But that solves only this instance of the issue. What if the user customized `imenu-generic-expression` so that every line with `5` in it is listed in the imenu index? I'd go for this instead:

```diff
diff --git a/package-lint.el b/package-lint.el
index 4732cff..521aec2 100644
--- a/package-lint.el
+++ b/package-lint.el
@@ -613,10 +613,13 @@ The returned list is of the form (SYMBOL-NAME . POSITION)."
   ;; enabled, Imenu will use its index anyway.
   (let ((result '())
         (index
-         ;; In case it's actually Semantic, tell it not to decorate symbol
-         ;; names.
          (save-excursion
-           (let ((semantic-imenu-summary-function 'semantic-format-tag-name))
+           ;; Use the default imenu expression list so that we're not confused
+           ;; by user customizations.
+           (let ((imenu-generic-expression lisp-imenu-generic-expression)
+                 ;; In case it's actually Semantic, tell it not to decorate
+                 ;; symbol names.
+                 (semantic-imenu-summary-function 'semantic-format-tag-name))
              (funcall imenu-create-index-function)))))
     (dolist (entry index)
       (pcase entry
```

Tested with outshine enabled, both with and without semantic, still works fine, still pulls the data from semantic if possible.

Fanael:
Committed the above. The problem should now be fixed, so closing the PR.

Fanael:
> The better solution may be to not rely on `imenu`, i.e. use `semantic` unconditionally.

Only if we move checking to a separate process (which, by the way, may be a good idea after all). I don't want semantic running in my Emacs.

purcell:
Nice, thanks. Agree with avoiding semantic if we can.

PR #58, opened 2017-03-30
Warn about literal references to Emacs user directory or files.

### LINKED ISSUE #24: Warn about literal references to ~/.emacs.d
A library should use `user-emacs-directory` for the path to the Emacs directory, and `locate-user-emacs-file` to find its data or config files within this directory.

I'll open a PR if I can find some time.

alphapapa:
Fixes #24.

### DIFF
--- a/package-lint.el
+++ b/package-lint.el
@@ -241,6 +241,7 @@ This is bound dynamically while the checks run.")
             (package-lint--check-lexical-binding-requires-emacs-24 deps)
             (package-lint--check-libraries-available-in-emacs deps)
             (package-lint--check-macros-functions-available-in-emacs deps))
+          (package-lint--check-for-literal-emacs-path)
           (let ((definitions (package-lint--get-defs)))
             (package-lint--check-defs-prefix definitions)
             (package-lint--check-symbol-separators definitions)))))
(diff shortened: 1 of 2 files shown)


alphapapa:
If I'm reading the logs correctly, it's failing because it's detecting itself...  LOL  :)

package-lint.el
```diff
@@ -253,6 +254,15 @@ This is bound dynamically while the checks run.")
 
 ;;; Checks
 
+(defun package-lint--check-for-literal-emacs-path ()
+  "Verify that package does not refer to \"~/.emacs.d\" literally.
+Instead it should use `user-emacs-directory' and `locate-user-emacs-file'."
+  (cl-loop
+   while (re-search-forward (rx (syntax string-quote) (1+ (not (syntax string-quote))) (or "/" "\\") ".emacs.d") nil t)
```
Fanael: Why `cl-loop` instead of just `while`?

alphapapa: Ah, good point, I whittled down the contents of the loop but didn't think to change it to `while`.

purcell:
> If I'm reading the logs correctly, it's failing because it's detecting itself... LOL :)

LOL indeed! You could cheat by using something like `\.emacs\.d`?

alphapapa:
Thanks, I think that should do it.  :)

Fanael:
Tests are missing.

alphapapa:
Guess I haven't been paying attention, I didn't think about that.  Looks simple, I will do that in the next day or two.  Thanks.

alphapapa:
Test added, seems to work.  Hope I did everything right.  :)

Fanael:
I amended it to fix the lint failure, merged as f41280e1b.

alphapapa:
@Fanael Sorry if I'm being dense, but as far as I can tell, https://github.com/purcell/package-lint/commit/f41280e1bc67a78eef0a3c28f8a73dc4521a5875 is exactly the same as https://github.com/purcell/package-lint/pull/58/commits/f8dcb552e52fa6e4538c791502ec64ee37421509, which is failing the CI, so the CI should fail with https://github.com/purcell/package-lint/commit/f41280e1bc67a78eef0a3c28f8a73dc4521a5875 also, right?

PR #61, opened 2017-04-11
Note that MELPA strictly requires hyphen separators

alphapapa:
Still seeing a lot of new recipe submissions ignoring this, which requires manually asking the author to fix it, which defeats the whole point of this check.  Maybe this will help.


purcell:
I don't want this to be about MELPA, because it's about supporting community standards. So I'd be happy to accept this if it said `(See Elisp Coding Conventions)` instead (and if the tests were updated to match).

alphapapa:
Understood.  And sorry that I forgot to update the tests.  :)

purcell:
GREAT - thanks

PR #63, opened 2017-04-16
Check autoloads

alphapapa:
This adds checks for common necessary and unnecessary autoloads, currently `define-minor-mode` and interactive commands.

### DIFF
--- a/package-lint-test.el
+++ b/package-lint-test.el
@@ -44,6 +44,13 @@ used."
     (let ((buffer-file-name "test.el"))
       (package-lint-buffer))))
 
+(ert-deftest package-lint-test-check-autoloads-on-private-functions ()
+  (should
+   (equal
+    '((3 0 warning "Private functions generally should not be autoloaded."))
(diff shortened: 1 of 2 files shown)


purcell:
Would you be able to split this out into 2 PRs, perhaps? I really like the check for autoloads on private functions, but I think the advice to add autoloads for all mode definitions and interactive functions is too broad, and likely to lead to messy packages. The guildelines are more nuanced, and I'm not yet sure if they can be automated.

(Further, I wonder if there's a case for adding "info"-level messages to `package-lint` for instances like this. Flycheck supports them IIRC, and we could avoid failing the `package-lint-buffer` check if only info-level messages are present. Opinions, @Fanael?)

alphapapa:
Sure, no problem.

I definitely would have used an `info`-level message if there were one.  In fact, I tried using that on one of the other PRs I sent recently, but found that it didn't work, so I had to use a warning.

alphapapa:
Submitted #64 and #65.  Closing this one for now; I'll open an issue about the info messages.

PR #64, opened 2017-04-17
Check for autoload cookies on private functions

alphapapa:


purcell:
Great, thanks.

package-lint.el
```diff
@@ -261,6 +262,21 @@ This is bound dynamically while the checks run.")
 
 ;;; Checks
 
+(defun package-lint--check-autoloads-on-private-functions ()
+  "Verify that private functions don't have autoload cookies."
+  (goto-char (point-min))
+  (while (re-search-forward (rx bol "(defun" (1+ space)
```
Fanael: This really should use the definitions from `package-lint--get-defs`.

purcell: Good catch, thanks for fixing this.

alphapapa: Thanks.

PR #65, opened 2017-04-17
Check for commonly needed autoloads

alphapapa:


package-lint.el
```diff
@@ -261,6 +262,36 @@ This is bound dynamically while the checks run.")
...
+    (unless (or (nth 3 (syntax-ppss))
+                (nth 4 (syntax-ppss)))
+      ;; Not in string or comment
+      (save-excursion
+        (when (re-search-backward (rx bol "(defun ") nil t)
+          (forward-line -1)
+          (unless (looking-at (rx ";;;###autoload"))
+            (package-lint--error
+             (line-number-at-pos) (current-column) 'warning
+             "Interactive commands generally should be autoloaded.")))))))
```
DamienCassou: I'm not native English speaker, but I would rather reverse "should" and "generally": "should generally be autoloaded."

package-lint.el
```diff
@@ -261,6 +262,36 @@ This is bound dynamically while the checks run.")
...
+          ;; Not in string or comment
+          (save-excursion
+            (forward-line -1)
+            (unless (looking-at (rx ";;;###autoload"))
+              (package-lint--error
+               (line-number-at-pos) (current-column) 'warning
+               (format "\"%s\" generally should be autoloaded." def))))))))
+  ;; Check interactive functions
+  (goto-char (point-min))
+  (while (re-search-forward (rx "(interactive)") nil t)
```
DamienCassou: why not `search-forward` if you don't need regular expression power?

DamienCassou:
This PR (which I like) has conflicts.

purcell:
Thanks for this.

I'm not sure how easy it is to come up with reasonable heuristics for suggesting autoloads. Many packages will use minor modes internally that are not meant to be used directly by users. It might be true to say that a package should generally have at least one autoload. Major modes are a more likely candidate than minor modes, generally. In any case, this is likely to be an "info"-level item of feedback rather than a warning.

package-lint.el
alphapapa: Good point, thanks.

package-lint.el
alphapapa: I guess this depends on how you feel about "split infinitives."  ;)  Personally, I think it's fine "to boldly go," while others would prefer "to go boldly".

alphapapa:
@DamienCassou It didn't when it was current, but by now... ;)

@purcell Yes, definitely more suitable for info than warning.  If you want, I can adjust that and resubmit, or if you think this PR isn't worth it, that's fine too.

package-lint-test.el
```diff
@@ -44,6 +44,22 @@ used."
...
+  ;; Positives
+  (should
+   (equal
+    '((2 0 warning "\"define-minor-mode\" generally should be autoloaded."))
+    (package-lint-test--run "(define-minor-mode test-something ")))
+  (should
+   (equal
+    '((2 0 warning "Interactive commands generally should be autoloaded."))
+    (package-lint-test--run "(defun test-something \n\"Docstring\"\n(interactive)\n")))
+  ;; Negatives
```
Fanael: The positive and negative testcases should be separate for consistency with other tests, see `package-lint-test-accept-*` vs `package-lint-test-{error,warn}-*`.

package-lint-test.el
```diff
@@ -44,6 +44,22 @@ used."
     (let ((buffer-file-name "test.el"))
       (package-lint-buffer))))
 
+(ert-deftest package-lint-test-check-for-needed-autoloads ()
+  ;; Positives
+  (should
+   (equal
+    '((2 0 warning "\"define-minor-mode\" generally should be autoloaded."))
+    (package-lint-test--run "(define-minor-mode test-something ")))
```
Fanael: Try to make the testcase source a valid Lisp source, like `(define-minor-mode test-something "")`.

package-lint-test.el
```diff
@@ -44,6 +44,22 @@ used."
...
+  (should
+   (equal
+    '((2 0 warning "Interactive commands generally should be autoloaded."))
+    (package-lint-test--run "(defun test-something \n\"Docstring\"\n(interactive)\n")))
```
Fanael: Again, an incomplete sexp.

package-lint-test.el
```diff
@@ -44,6 +44,22 @@ used."
...
+  ;; Negatives
+  (should
+   (equal nil (package-lint-test--run ";;;###autoload\n(define-minor-mode test-something ")))
```
Fanael: Incomplete sexp.

package-lint-test.el
```diff
@@ -44,6 +44,22 @@ used."
...
+  (should
+   (equal nil (package-lint-test--run ";;;###autoload\n(defun test-something \n\"Docstring\"\n(interactive)\n"))))
```
Fanael: Incomplete sexp.

package-lint.el
```diff
@@ -261,6 +262,36 @@ This is bound dynamically while the checks run.")
 
 ;;; Checks
 
+(defun package-lint--check-for-needed-autoloads ()
+  "Verify that autoload cookies are on definitions that generally need them."
+  ;; Check common defs
+  (let ((defs '("define-minor-mode")))
+    (dolist (def defs)
```
Fanael: I'd like to avoid dynamically constructing regexps wherever possible, so get rid of this loop and use a single regexp: `(rx bol "(" (group (or "define-minor-mode" "define-foo" "define-bar")) (1+ space))`, then extract the match using `match-string-no-properties`…

package-lint.el
```diff
@@ -261,6 +262,36 @@ This is bound dynamically while the checks run.")
...
+      (while (re-search-forward (rx-to-string `(seq bol "(" ,def (1+ space))) nil t)
+        (unless (or (nth 3 (syntax-ppss))
+                    (nth 4 (syntax-ppss)))
+          ;; Not in string or comment
+          (save-excursion
+            (forward-line -1)
+            (unless (looking-at (rx ";;;###autoload"))
+              (package-lint--error
+               (line-number-at-pos) (current-column) 'warning
+               (format "\"%s\" generally should be autoloaded." def))))))))
```
Fanael: …here.

package-lint.el
```diff
@@ -261,6 +262,36 @@ This is bound dynamically while the checks run.")
...
+  (let ((defs '("define-minor-mode")))
+    (dolist (def defs)
+      (goto-char (point-min))
+      (while (re-search-forward (rx-to-string `(seq bol "(" ,def (1+ space))) nil t)
+        (unless (or (nth 3 (syntax-ppss))
+                    (nth 4 (syntax-ppss)))
+          ;; Not in string or comment
+          (save-excursion
+            (forward-line -1)
+            (unless (looking-at (rx ";;;###autoload"))
```
Fanael: (this then needs to be `looking-at-p` so it doesn't modify match data)

Fanael (changes requested):

Fanael:
> Yes, definitely more suitable for info than warning. If you want, I can adjust that and resubmit

Yeah, I'm in favor of making it info-level. If you rebase it on top of master and address the review feedback, I'll merge it, unless @purcell objects.

> I'm not sure how easy it is to come up with reasonable heuristics for suggesting autoloads

Sure, but it's a relatively reasonable starting point IMO, and we can always tune it down later if it proves to be too annoying.

alphapapa:
@Fanael Thanks for the feedback, I'll rebase and make those changes soon.

Fanael:
Is this PR still alive?

joewreschnig:
> It might be true to say that a package should generally have at least one autoload. 

I think most library packages (e.g. dash, f, seq) do not have any autoloads.

A decent guideline might be:
- Major or minor modes without "--" in their names must be autoloaded.
- If you have any `interactive` forms, you must also have at least one autoload (not necessarily the same as those interactive commands

alphapapa:
Sorry, I'm busy with some other things right now.  Maybe someone else could take this PR forward, or if not, it's probably not that important.  Thanks.

PR #70, opened 2017-06-06
Report checkdoc errors too

DamienCassou:
This adds checkdoc to the list of checks. The only problem I found is that `package-lint.el` has checkdoc warnings so this PR won't build. The problematic documentation is:

```
(defun package-lint--check-version-regexp-list (valid-deps list rx-start rx-end)
  "Warn about matches of REGEXP when VERSION is not in VALID-DEPS.
LIST is an alist of (VERSION . REGEXP*).
REGEXP is (concat RX-START REGEXP* RX-END) for each REGEXP*."
...)
```

checkdoc reports that `"matches"` should probably be `"match"` instead. I don't know what you want to do about that.

### DIFF
--- a/package-lint-test.el
+++ b/package-lint-test.el
@@ -41,6 +41,7 @@ it's nil, the default is used."
     (insert (or version ";; Package-Version: 0\n"))
     (insert (or url ";; URL: https://package-lint.test/p?q#f\n"))
     (insert (or commentary ";;; Commentary:\n;; A test package, for testing.\n"))
+    (insert "\n;;; Code:\n")
     (insert contents)
     (insert "\n" (or provide "(provide 'test)\n"))
     (insert (or footer "\n\n;;; test.el ends here\n"))
@@ -103,7 +104,7 @@ it's nil, the default is used."
     (should-not (package-lint-test--run "(defcustom test-something (kbd \"C-x <ESC> <ESC>\"))"))))
 
 (ert-deftest package-lint-test-error-autoloads-on-private-functions ()
-  (should (equal '() (package-lint-test--run "(defun test--private-function ())")))
+  (should (equal '() (package-lint-test--run "(defun test--private-function () \"Do something.\")")))
   (should
    (equal
     '((6 0 warning "Private functions generally should not be autoloaded."))
@@ -202,8 +203,8 @@ it's nil, the default is used."
 
 (ert-deftest package-lint-test-warn-empty-summary ()
   (should
-   (equal
-    '((1 1 warning "Package should have a non-empty summary."))
+   (member
+    '(1 1 warning "Package should have a non-empty summary.")
     (package-lint-test--run "" ";;; test.el ---\n"))))
 
 (ert-deftest package-lint-test-warn-too-long-summary ()
@@ -305,9 +306,9 @@ Alternatively, depend on (emacs \"24.3\") or greater, in which cl-lib is bundled
     '((6 1 error "`test-thing/bar' contains a non-standard separator `/', use hyphens instead (see Elisp Coding Conventions).")
       (7 1 error "`test-thing:bar' contains a non-standard separator `:', use hyphens instead (see Elisp Coding Conventions)."))
     (package-lint-test--run
-     "(defun test-thing/bar () t)\n(defun test-thing:bar () nil)")))
+     "(defun test-thing/bar () \"Do something.\" t)\n(defun test-thing:bar () \"Do something.\" nil)")))
   ;; But accept /= when at the end.
-  (should (equal '() (package-lint-test--run "(defun test-/= (a b) t)"))))
+  (should (equal '() (package-lint-test--run "(defun test-/= (a b) \"Do something with A and B.\" t)"))))
 
 (ert-deftest package-lint-test-error-unprefixed-definitions ()
   (should
@@ -321,7 +322,7 @@ Alternatively, depend on (emacs \"24.3\") or greater, in which cl-lib is bundled
 
 (ert-deftest package-lint-test-accept-prefixed-definitions ()
   (should (equal '() (package-lint-test--run
-                      "(defun test-foo ())\n(defun test ())")))
+                      "(defun test-foo () \"Do something.\")\n(defun test () \"Do something.\")")))
   (should (equal '() (package-lint-test--run
                       "(define-globalized-minor-mode global-test-mode ignore ignore :require 'test)")))
   (should (equal '() (package-lint-test--run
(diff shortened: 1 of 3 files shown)


DamienCassou:
There is also a problem with the new way of quoting:

```
Test package-lint-test-warning-checkdoc condition:
    (ert-test-failed
     ((should
       (equal '...
	(package-lint-test--run "(defun test-foo (bar) \"Not mentioning parameter.\")")))
      :form
      (equal
       ((5 22 warning "Argument ‘bar’ should appear (as BAR) in the doc string"))
       ((5 22 warning "Argument `bar' should appear (as BAR) in the doc string")))
```
The error message is different depending on the Emacs version. I think this could be solved by using a regular expression to match against the error message (with `pcase`?).

Fanael:
> The error message is different depending on the Emacs version. I think this could be solved by using a regular expression to match against the error message (with `pcase`?).

No need, simply bind `text-quoting-style` to `'grave`.

package-lint.el
```diff
@@ -295,6 +297,19 @@ Instead it should use `user-emacs-directory' or `locate-user-emacs-file'."
          line-no 1 'warning
          (format "You should include standard keywords: see the variable `finder-known-keywords'."))))))
 
+(defun package-lint--check-documentation ()
+  "Verify that package keywords are listed in `finder-known-keywords'."
+  (cl-letf (((symbol-function #'checkdoc-error) (lambda (point msg)
+                                                  (when point
```
Fanael: Move that lambda to the next line. Indentation this deep makes me uncomfortable.

DamienCassou:
> No need, simply bind text-quoting-style to 'grave around the package-lint-buffer call.

done. But I find it weird not to respect this user setting just to simplify testing.

> Move that lambda to the next line

done

Fanael:
Myself, I don't see much point in this, because all it does is call checkdoc, which:
1. everyone uses anyway,
2. you get duplicate messages if you're using flycheck together with flycheck-package.

But I'll defer to @purcell's judgment on this one.

package-lint.el
```diff
@@ -295,6 +297,22 @@ Instead it should use `user-emacs-directory' or `locate-user-emacs-file'."
...
+               (when point
+                 (package-lint--error
+                  (line-number-at-pos point)
+                  (save-excursion
+                    (goto-char point)
+                    (current-column))
+                  'warning
+                  msg)))))
+    ;; ensure same error message on all supported Emacs version:
+    (let ((text-quoting-style nil))
```
Fanael: Not here. In the *testsuite*.

package-lint-test.el
```diff
@@ -38,45 +38,46 @@ used."
     (emacs-lisp-mode)
     (insert (or header ";;; test.el --- A test\n"))
     (insert (or version ";; Package-Version: 0\n"))
+    (insert "\n;;; Code:\n")
     (insert contents)
     (insert "\n" (or provide "(provide 'test)\n"))
     (insert (or footer "\n\n;;; test.el ends here\n"))
     (let ((buffer-file-name "test.el"))
       (package-lint-buffer))))
```
Fanael: Bind it *here*.

Fanael:
> done. But I find it weird not to respect this user setting just to simplify testing.

I meant in the testsuite, where we're in control of user settings anyway. The linter proper should follow user settings where sensible.

package-lint.el
DamienCassou: sorry, I just woke up! Thanks.

package-lint-test.el
DamienCassou: done

DamienCassou:
> I meant in the testsuite, where we're in control of user settings anyway. The linter proper should follow user settings where sensible.

obviously, sorry :-).

purcell:
Thanks. This is helpful in terms of linting packages, but it has the issue that the checking is redundant when used via `flycheck-package`. So I'm not sure if we should really incorporate this.

I'd prefer to first tackle checks like "has a commentary section which is not empty".

DamienCassou:
Is `flycheck-package` runnable from a continuous integration server? I don't see how to configure that in its README file (I have to admit I don't really understand the purpose of this package at all, the only sentence describing it being "This library provides a flycheck checker for the metadata in Emacs Lisp files which are intended to be packages").

purcell:
> Is flycheck-package runnable from a continuous integration server? I don't see how to configure that in its README file (I have to admit I don't really understand the purpose of this package at all, the only sentence describing it being "This library provides a flycheck checker for the metadata in Emacs Lisp files which are intended to be packages").

No, it's a `flycheck` checker. i.e. you enable `flycheck`, and it shows you - directly in the buffer - any messages produced by running `package-lint` on the buffer.

DamienCassou:
Thanks for the explanation. IIUC, both of you are saying that because flycheck can already highlight checkdoc errors, you don't see the point of adding it to package-lint. Am I right?

What I need is checking in a continuous integration server. I would like checkdoc to run in this context (others are doing it too, e.g., https://github.com/NicolasPetton/Indium/blob/master/run-lint.el). I could do it manually but I thought it would be nicer that package-lint does it for me (also, I wouldn't have to copy/paste my configuration to all my projects and `package-lint` has that in its todo list).

purcell:
> Thanks for the explanation. IIUC, both of you are saying that because flycheck can already highlight checkdoc errors, you don't see the point of adding it to package-lint. Am I right?

Yes, but I understand the contradiction. I agree it would be nice to make `package-lint` self-contained, and *if* we consider `checkdoc` violations to be a packaging issue (which I'm undecided on, really), then it would be useful to include `checkdoc` feedback in `package-lint`'s summary. Given how `flycheck-package` currently works, I don't want to introduce duplicate work. So options might be to selectively disable the `checkdoc` portion of `package-lint`'s checking when run from `flycheck-package`. Not sure... I need to think about it a bit.

Fanael:
> I'd prefer to first tackle checks like "has a commentary section which is not empty".

Done, 548eaa619c669dd990023be39a52b933b875c9b7.

DamienCassou:
I solved the conflicts on this branch introduced by 548eaa619c669dd990023be39a52b933b875c9b7.

DamienCassou:
any news about this PR?

purcell:
> any news about this PR?

It's currently not applying cleanly. But as noted above, I'm still not 100% that it's reasonable to consider `checkdoc` issues packaging errors. To avoid noise, I think I'd probably be happier if we simply ran "checkdoc" and then generated a single warning on the first or last line indicating that checkdoc is unhappy.

DamienCassou:
Fixed.

purcell:
This is great, thanks!

The patch causes `package-lint` to fail when checking itself, which is why the CI build failed. Would you mind fixing that too please?

New package authors may not know what checkdoc is, so may I also suggest that the new error message end with the text ": run `M-x checkdoc' to fix" ?

DamienCassou:
> The patch causes package-lint to fail when checking itself, which is why the CI build failed. Would you mind fixing that too please?

I talked about that in the very [first message here](https://github.com/purcell/package-lint/pull/70#issue-234010730)

> may I also suggest that the new error message end with the text ": run `M-x checkdoc' to fix" ?

done

purcell:
Hmm, I fixed that and re-ran the build, but it's still failing on one test.

Fanael:
I restarted the build on Travis, now it apparently fails on three tests.

DamienCassou:
I've failed to find enough interest in this PR to get it fully working. You may close it.

PR #71, opened 2017-06-06
Add Emacs 25.2 build target

DamienCassou:


purcell:
Great, thanks for this!

PR #79, opened 2017-08-08
Test for reserved keybindings

alphapapa:
Been seeing a lot of package reviews with Steve having to tell people not to use reserved keybindings.  Maybe this will help.  :)

This may need some polishing, adjusting of the message, etc, but it seems to work pretty well.  The rx macro is nice!

### DIFF
--- a/package-lint-test.el
+++ b/package-lint-test.el
@@ -44,6 +44,94 @@ with the passed string, when it's nil, the default is used."
     (let ((buffer-file-name "test.el"))
       (package-lint-buffer))))
 
+(ert-deftest package-lint-test-reserved-keybindings ()
+  ;; C-c and a letter (either upper or lower case)
+  (should
+   (equal
+    '((5 13 warning "This key binding is reserved."))
+    (package-lint-test--run "(kbd \"C-c n\")")))
+  (should
+   (equal
+    '((5 7 warning "This key binding is reserved."))
+    (package-lint-test--run "\"\\C-cn\"")))
+  ;; C-c followed by a control character or a digit
+  (should
+   (equal
+    '((5 39 warning "This key binding is reserved."))
+    (package-lint-test--run "(defcustom test-something (kbd \"C-c 1\")")))
+  (should
+   (equal
+    '((5 7 warning "This key binding is reserved."))
+    (package-lint-test--run "\"\\C-c1\"")))
+  ;; C-c followed by {, }, <, >, : or ;
+  (should
+   (equal
+    '((5 39 warning "This key binding is reserved."))
+    (package-lint-test--run "(defcustom test-something (kbd \"C-c <\")")))
+  (should
+   (equal
+    '((5 7 warning "This key binding is reserved."))
+    (package-lint-test--run "\"\\C-c<\"")))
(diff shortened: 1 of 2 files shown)


purcell:
Ah! Great idea, and this is awesome. Thanks!

I wonder if it would make sense to generalise this, e.g. scan out all of the keybindings mentioned in the file and `eval` them to get the actual key values, rather than parsing variously-formatted "kbd" macro strings etc.

Also, the message could perhaps cite a reference, e.g. by mentioning the manual section title.

Fanael:
> I wonder if it would make sense to generalise this, e.g. scan out all of the keybindings mentioned in the file and `eval` them to get the actual key values, rather than parsing variously-formatted "kbd" macro strings etc.

Yeah, and I'm going to NAK this PR the way it's currently implemented — getting the data by evaluating `kbd` is easy enough and *much* less brittle than trying to parse it using regexes, so that's how it should be done. I believe we already have too many regex hacks in package-lint.

alphapapa:
> Also, the message could perhaps cite a reference, e.g. by mentioning the manual section title.

That's a good idea.  I figured the URL to that page on the GNU site was too long.

> I wonder if it would make sense to generalise this, e.g. scan out all of the keybindings mentioned in the file and eval them to get the actual key values, rather than parsing variously-formatted "kbd" macro strings etc.

I briefly considered that but thought I should avoid `eval`.  

> Yeah, and I'm going to NAK this PR the way it's currently implemented — getting the data by evaluating kbd is easy enough and much less brittle than trying to parse it using regexes, so that's how it should be done. I believe we already have too many regex hacks in package-lint.

I thought the `rx` form turned out okay, but okay.

purcell:
It's a nice `rx` form, nothing against that. :-)

alphapapa:
Not all key bindings use `kbd` anyway, so I guess you'll have to check every string in every function that can bind a key.

Fanael:
> I briefly considered that but thought I should avoid `eval`.

In general, that's certainly a valid approach, but in this case, i.e. the `kbd` function itself and a handful of other function which use it implicitly when passed a string, there's nothing wrong with just evaluting the kbd form. Furthermore, `kbd` is pure, so should be safe to `eval`.

> I thought the `rx` form turned out okay, but okay.

It looks pretty and handles the cases it was made for reasonably well, but consider the limitations of regexes you've discovered here, like
```emacs-lisp
;; TODO: Don't bind a key sequence ending in <ESC> except following another <ESC>
;; Can't find a way to make this work in the regexp.  Could test the resulting string but...
```
and
```emacs-lisp
;; FIXME: This is ugly but I can't find a way to negate a string at once.
;; And, anyway, it doesn't work.  :(
```

The `eval` approach should handle both with ease.

> Not all key bindings use `kbd` anyway, so I guess you'll have to check every string in every function that can bind a key.

There aren't that many of them, are there? There's `global-set-key`, `local-set-key`, `define-key`, any others?

Fanael:
By the way: https://www.gnu.org/software/emacs/manual/html_node/elisp/Input-Events.html should prove useful for trying to make sense of the values returned by `kbd`.

PR #80, opened 2017-08-10
Test for reserved keybindings

alphapapa:
cf. #79

### DIFF
--- a/package-lint-test.el
+++ b/package-lint-test.el
@@ -44,6 +44,65 @@ with the passed string, when it's nil, the default is used."
     (let ((buffer-file-name "test.el"))
       (package-lint-buffer))))
 
+(ert-deftest package-lint-test-reserved-keybindings ()
+  ;; C-c and a letter (either upper or lower case)
+  (should (equal (package-lint-test--run "(kbd \"C-c n\")")
+                 '((5 13 warning "Key sequences that begin with C-c are reserved (unless followed by another modifier sequence)"))))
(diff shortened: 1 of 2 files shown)


Fanael:
> FIXME: This also catches "C-c C-x", which is allowable, but I don't know how to get the modifier of the C-x.

Call `event-modifier` on `(aref lks 1)`.

alphapapa:
Thanks, it seems to be working now.

purcell:
Cool! It's not clear to me why the build is still failing, though?

alphapapa:
> Cool! It's not clear to me why the build is still failing, though?

Me neither.  The tests all pass when I run `ert-run-tests-interactively`.  And some of the other commits I pushed like this passed the tests, except on Emacs 24.1 and 24.2.

purcell:
I tracked down the problem. Bear with me.

purcell:
The code wasn't really doing the right thing when there was a `(kbd "...")` inside a `(define-key` or similar. I've simplified it and made it a bit more robust, and also changed the message so that it is the same for all cases, and simply refers people to the corresponding manual section.

alphapapa:
Thanks.

alphapapa:
Thanks.  I made so many mistakes in this code that I'm not sure it saved you any time over writing it yourself.

purcell:
> Thanks. I made so many mistakes in this code that I'm not sure it saved you any time over writing it yourself.

LOL, that's not true - I wouldn't have got round to writing it without you starting the ball rolling.

Issue #77, opened 2017-08-08
Detect missing :require on global minor modes


joewreschnig:
> [22.3.3 Defining Minor Modes](https://www.gnu.org/software/emacs/manual/html_node/elisp/Defining-Minor-Modes.html)
> One of the effects of making a minor mode global is that the _mode_ variable becomes a customization variable... For the saved variable to work, you should ensure that the `define-minor-mode` form is evaluated each time Emacs starts; for packages that are not part of Emacs, the easiest way to do this is to specify a `:require` keyword.

`package-lint` should warn if `:require` is missing, and probably also if its value doesn't match the package/feature name.

This is also the case for `define-globalized-minor-mode`.

PR #82, opened 2017-08-17
Check that global minor modes are defined properly.

joewreschnig:
Global minor modes should be defined with
`define-globalized-minor-mode`, and must `:require` the corresponding
feature so setting the automatically-created customization variable
behaves properly.

These checks use a new helper, `package-lint–-check-objects-by-regexp`,
which reads Lisp objects out of the buffer and passes them to a
function specific to that form.

Test cases for the new checks have been added, and old test data
updated to avoid triggering the new checks.

This addresses issue #77.


joewreschnig:
Another helper function introduced in this PR, `package-lint--error-at-point`, could also simplify a number of other checks. Switching those is not included in this PR.

purcell:
Thanks Joe - nice work. That helper looks handy too.

Issue #78, opened 2017-08-08
Warn against ending customization groups with `-mode'


joewreschnig:
Customization groups for a mode should only end in `-mode` if they are subgroups of the same name without `-mode` (e.g. `diff` and `diff-mode` or `abbrev` and `abbrev-mode`).

PR #84, opened 2017-08-19
Detect customization group names ending in “-mode”.

joewreschnig:
Such names are not idiomatic, and at odds with the default behavior of
e.g. `define-minor-mode`. They are only used when the group name
without “-mode” is already used for non-mode-specific configuration
options, in which case it should be used as the parent group.

This addresses issue #78.


purcell:
Awesome, thanks.

Issue #67, opened 2017-04-17
The check against literal references to .emacs.d is a bit overzealous


Fanael:
The check warns on:
```emacs-lisp
(defvar foo "/foo/foo.emacs.dat")
```
even though this is, *obviously*, not a reference to `.emacs.d`.

Any idea how to make this check more bulletproof?

PR #85, opened 2017-08-20
Do not warn about paths like “/foo/foo.emacs.dat”.

joewreschnig:
Instead, bad paths must have a symbol delimiter (e.g. a quote) or a /
before the first dot, and a word break after the “d”.

Update the existing test cases so they are valid Emacs Lisp, by
making the test path a string.

This addresses issue #67.

### DIFF
--- a/package-lint-test.el
+++ b/package-lint-test.el
@@ -120,8 +120,13 @@ with the passed string, when it's nil, the default is used."
 (ert-deftest package-lint-test-warn-literal-emacs-path ()
   (should
    (equal
-    '((5 8 warning "Use variable `user-emacs-directory' or function `locate-user-emacs-file' instead of a literal path to the Emacs user directory or files."))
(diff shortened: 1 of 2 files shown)


purcell:
Great, thanks!

purcell:
Joe, I've sent you an invitation for you to join as a collaborator on this repo. Feel free to make little fixes like this directly, or continue to file as PRs if you'd like reviews or second opinions. :-)

joewreschnig:
Thanks! There's more things in the README I'd like to knock off. I'll stick to PRs for new checks, but I expect I'll self-merge if there's no objection after a couple days.

purcell:
Sounds good, thanks!

alphapapa:
@joewreschnig I just realized that you are the original author of Mutagen.  Nice to meet you here in Emacsland.  Thanks for Mutagen, even though you don't work on it anymore.  It's an integral part of my music library scripts.

PR #86, opened 2017-08-21
Check that packages have a valid HTTPS or HTTP URL.

joewreschnig:
This requires +1’ing most test case line numbers.

### DIFF
--- a/package-lint-test.el
+++ b/package-lint-test.el
@@ -4,6 +4,7 @@
 
 ;; Author: Steve Purcell <steve@sanityinc.com>
 ;;         Fanael Linithien <fanael4@gmail.com>
+;; URL: https://github.com/purcell/package-lint
 ;; Version: 0
 
 ;; This program is free software; you can redistribute it and/or modify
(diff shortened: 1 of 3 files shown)


joewreschnig:
Although almost all the packages I checked seem to be using `URL` (or nothing), [the official Emacs documentation suggests `Homepage`](https://www.gnu.org/software/emacs/manual/html_node/elisp/Library-Headers.html). The latter now sounds a little anachronistic to me, though I understand why it might be preferred since "URL" is vague.

package-lint-test.el
```diff
@@ -27,16 +28,18 @@
...
-with the passed string, when it's nil, the default is used."
+HEADER, VERSION, FOOTER, PROVIDE, COMMENTARY, and URL can be
+either strings or nil; when one is a string, the corresponding
+package boilerplate part is replaced with the passed string, when
+it's nil, the default is used."
   (with-temp-buffer
     (emacs-lisp-mode)
     (insert (or header ";;; test.el --- A test\n"))
     (insert (or version ";; Package-Version: 0\n"))
+    (insert (or url ";; URL: https://package-lint.test/p?q#f\n"))
```
Fanael: Use `example.com` instead, so we can be sure the URL never points to anything nasty.

joewreschnig: [`.test` is reserved in the same way for something closer to the purpose at hand.](https://tools.ietf.org/html/rfc6761)

PR #87, opened 2017-08-22
“.emacs” (or “~/.emacs.d”, etc.) is not redundant in summaries.

joewreschnig:


package-lint-test.el
```diff
@@ -208,6 +208,9 @@ with the passed string, when it's nil, the default is used."
 (ert-deftest package-lint-test-accept-emacs-lisp-in-summary ()
   (should (equal '() (package-lint-test--run "" ";;; test.el --- Emacs Lisp test framework\n"))))
 
+(ert-deftest package-lint-test-accept-.emacs-in-summary ()
+  (should (equal '() (package-lint-test--run "" ";;; test.el --- Something for .emacs\n"))))
```
alphapapa: This is not specific to this commit, but I'm curious: is there a reason we use `(should (equal '() (package-lint-test--run...` instead of `(should-not (package-lint-test--run...`?

joewreschnig: I would use `should-not` to check predicates, not values.

We are interested in knowing that no errors were generated, not that `package-lint-test--run` returned a false value. That nil, the empty list, and falsity are the same is an implementation detail (some might say mistake) of Lisp.

Other unit testing libraries often offer assertions like `should-equal` or `should-be-empty` which do an even better job of indicating this kind of intent, as well as generate more readable failure messages.

PR #92, opened 2017-10-06
"Homepage" is a valid url header keywords 

tarsius:
Please see the commit messages.

### DIFF
--- a/package-lint.el
+++ b/package-lint.el
@@ -357,8 +357,8 @@ Instead it should use `user-emacs-directory' or `locate-user-emacs-file'."
          (format "You should include standard keywords: see the variable `finder-known-keywords'."))))))
 
 (defun package-lint--check-url-header ()
-  "Verify that the package has an HTTPS or HTTP URL header."
-  (if (package-lint--goto-header "URL")
+  "Verify that the package has an HTTPS or HTTP Homepage/URL header."
+  (if (package-lint--goto-header "\\(?:URL\\|Homepage\\)")
       (let ((url (match-string 3))
             (url-start (match-beginning 3)))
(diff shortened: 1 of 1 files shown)


joewreschnig:
I don't mind "Homepage" but I don't think we should permit "X-..." headers, if the point of package-lint is to enforce best practices.

tarsius:
Yeah, I did that in a separate commit for easy dropability.

tarsius:
Ah, and it seems I have to update the tests. Will do that after @purcell has decided what to do about `X-*`.

purcell:
Yeah, I think I agree with @joewreschnig here that we should disallow `X-*`. If @tarsius has stats indicating that `X-URL` is a de-facto standard, I'm happy to let that slide though.

tarsius:
No stats this time. And I would prefer if maintainers dropped the unnecessary `X-` too.

Fanael:
I agree that we should recognize `Homepage`, but I'm going to reject this PR because the check should be rewritten to simply use `lm-homepage` instead.

Fanael:
Whoops, `lm-homepage` is not defined in Emacs 24.3 and below.

Fanael:
> I don't mind "Homepage" but I don't think we should permit "X-..." headers, if the point of package-lint is to enforce best practices.

Thirded.

> Ah, and it seems I have to update the tests.

Yeah, if you do that it'll be ready to merge.

Fanael:
Actually, don't bother, I'll do it: rebased and amended as ff64e1171e8.

purcell:
Thanks all!

tarsius:
Thanks!

PR #95, opened 2017-11-03
Add package-lint binary

### LINKED ISSUE #73: Get data from Cask
I currently have this in my [Makefile](https://github.com/DamienCassou/klassified.el/blob/master/Makefile) to run package-lint:

```makefile
lint : $(SRCS) clean-elc
	# Run package-lint to check for packaging mistakes
	${CASK} emacs $(EMACSFLAGS) \
	--eval "(require 'package)" \
	--eval "(push '(\"melpa\" . \"http://melpa.org/packages/\") package-archives)" \
	--eval "(package-initialize)" \
	--eval "(package-refresh-contents)" \
	-l package-lint.el \
	-f package-lint-batch-and-exit klassified.el

```

I see several problems here and would like to simplify that:

1. I don't want to specify `package-archives` here as I already specify it in my Cask file
    ```lisp
   (source gnu)
   (source melpa)
   ```
1. I don't want to run `package-initialize` and `package-refresh-contents` manually. This should be the responsibility of the package-lint function I call (e.g., `package-lint-batch-and-exit`).

Basically, I would like to just write:

```Makefile
lint : $(SRCS) clean-elc
	# Run package-lint to check for packaging mistakes
	${CASK} emacs $(EMACSFLAGS) \
	-l package-lint.el \
	-f package-lint-FROM-CASK-batch-and-exit klassified.el
```

Please note the `FROM-CASK` above to tell `package-lint` to use the Cask specification. If you think package-lint is not appropriate for that, what do you think about a `package-lint-cask` responsible for calling `package-lint` properly (in the same way we have `flycheck-cask`).

Fuco1:
For projects using cask there is the handy `cask exec` command to execute "binaries" packaged with packages.  They are stored in a `bin` directory as simple executable files.

This patch adds a `package-lint` binary so that you can simply `cask exec package-lint` and it will automatically set up the load path, archives and everything according to `Cask`.  You can also specify the files you want to check as arguments (by default uses all files returned by `cask files`).

This should fix #73 and also make builds on CIs much more pleasant.

I can imagine this to also be used in travis-ci for MELPA if we can detect that the project uses Cask.  We can later make it more configurable to be able to run without cask as well... but I would strongly push for cask adoption among "serious" packages (maybe even require it by package-lint itself as a check).

### DIFF
--- a/bin/package-lint
+++ b/bin/package-lint
@@ -1,0 +1,63 @@
+#!/usr/bin/env bash
+
+usage () {
+    cat <<EOF
+[cask exec] package-lint [FILES]
+
+Arguments:
+
+FILES:
+  Package lint will use files defined in Cask by default if no files are
+  specified as arguments.
+
+EOF
+}
+
+CASK="${CASK:=cask}"
+EMACS="${EMACS:=emacs}"
+FILES="$*"
+
+INIT_PACKAGE_EL="(progn
+  (require 'package)
+  (setq package-user-dir \"$($CASK package-directory)\")
+  (setq package-archives
+        (mapcar
+         (lambda (archive)
+           (list archive))
+         (cl-delete-if
+          (lambda (file) (string-prefix-p \".\" file))
+          (directory-files (concat package-user-dir \"/archives\")))))
+  (package-read-all-archive-contents))"
+
+if [ -z "$FILES" ]; then
+    FILES=$("$CASK" files)
+fi
+
+EMACS_ARGS=()
+while [[ "$#" -gt 0 ]]
+do
+    case "$1" in
+        "-h"|"--help")
+            usage
+            exit
+            ;;
+        "-L"|"--directory"|"-f"|"--funcall"|"-l"|"--load"|"--eval"|"--execute")
+            EMACS_ARGS+=("$1")
+            EMACS_ARGS+=("$2")
+            shift
+            shift
+            ;;
+        *)
+            shift
+            ;;
+    esac
+done
+
+
+"$EMACS" -Q -batch \
+         "${EMACS_ARGS[@]}" \
+         --eval "$INIT_PACKAGE_EL" \
+         -L . \
+         -l package-lint.el \
+         -f package-lint-batch-and-exit \
+         $FILES


Fuco1:
cc @DamienCassou you might be interested in this.

purcell:
This is great! If you have a few moments, it would be great if you could change the `travis` config of this repo to use the script on itself, and also mention the script in the README.

Fuco1:
@purcell I can do that but I would have to add a `Cask` file to the project as this depends on cask for now (as it wouldn't install through package.el with the binary anyway I think?).  Is that ok?

conao3:
This PR is reasonable! Cask integration is useful for many Elisp developers!
How this PR state?

Fuco1:
I still use it for my own development.

purcell:
I do like this. What's held me back from merging is the need for `cask`, and compatibility with older Emacs versions.

With the current version of this:
- cask is required, but the usage text suggests it might not be
- package-lint's dependencies won't be installed if not present, e.g. `cl-lib` on Emacs < 24.4, and we currently target compatibility with all versions >= 24.1

I'm looking at writing a cask-less version of this, for these reasons. We've done some stuff along these lines in MELPA, for the sandbox loading support.

Fuco1:
There's also the [emake](https://github.com/vermiculus/emake.el) project which supports many linters for elisp including package lint.  Might be the easiest way to run.  It has no dependencies other than `make` and `emacs`

conao3:
I may be misunderstood.
Cask is used by people who want to use it (in an environment that can do it), and those who cannot use it are just not using it.
In other words, it can be used for developing packages for new Emacs without bothering existing users.

For example, I have written (and maintain) a package that supports Emacs-23, but this binary will not work for it. But that's fine. Because there is little to consider such old Emacs.

ericdallo:
Any news on that?

alphapapa:
https://github.com/alphapapa/makem.sh is another very simple way to run `package-lint`.

Fuco1:
I think the only reason for hard-wiring cask here is the call to cask files.  I think this should just be removed and files should be passed externally, either piped in (`cask files | ...`) or somehow provided manually.

I'm using Eask more these days but on my other projects all the binaries are actually compatible with both, because they just start Emacs and load the script, whereas cask and/or Eask provide environment (load path etc) through environment.

So I think if we make this a shell script which only depends on Emacs and not calling cask or anything else, we can just include it and whatever else mechanism can be used to supply it with correct environment.

(btw, the amount of dev tools for Emacs is horrifying :D ... but I really don't like the Make-based ones :/)

PR #101, opened 2018-03-23
Don't display a buffer if there are no issues

tarsius:
Instead just show the message "No issues found" in the echo area.


purcell:
Great, thanks.

PR #102, opened 2018-03-29
Make it possible to force acceptance of a dependency version

DamienCassou:
Add a variable `package-lint-accepted-dependencies` to list dependencies that should be accepted even if not in any package archive.

This PR also adds a test for an existing check.

### DIFF
--- a/package-lint-test.el
+++ b/package-lint-test.el
@@ -304,6 +304,13 @@ it's nil, the default is used."
     (should
      (equal '() (package-lint-test--run ";; Package-Requires: ((package-lint-foobar \"0.3.0\"))")))))
 
+(ert-deftest package-lint-test-dont-warn-dependency-too-high-4 ()
+  (let ((package-lint-accepted-dependencies '((package-lint-foobar (0 6 0))))
+        (package-archive-contents nil))
+    (package-lint-test-add-package-lint-foobar-to-archive '(0 5 0))
+    (should
+     (equal '() (package-lint-test--run ";; Package-Requires: ((package-lint-foobar \"0.6.0\"))")))))
+
 (ert-deftest package-lint-test-error-cl-lib-1.0-dep ()
   (should
    (member
--- a/package-lint.el
+++ b/package-lint.el
@@ -438,6 +438,31 @@ the form (PACKAGE-NAME PACKAGE-VERSION LINE-NO LINE-BEGINNING-OFFSET)."
           (format "Expected (package-name \"version-num\"), but found %S." entry)))))
     valid-deps))
 
+(defvar package-lint-accepted-dependencies nil
+  "List of pairs (PACKAGE-NAME VERSION) for which package-lint won't complain.
+
+If the checked file requires a dependency whose name and version
+are in this list, package-lint won't check the package-archives
+for the availability of the dependency's version.  This is useful
+for package authors who release several inter-dependent packages
+at the same time: an author won't have to first release the
+depended-upon package, wait for the package archive to be
+updated, and then release the depending package.
+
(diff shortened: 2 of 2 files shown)


purcell:
Thanks! The fixes here are useful. I'm not sure that the main change belongs in `package-lint`, though: it's quite an edge case, and authors could instead add fake entries to `package-alist` for their unpublished packages, couldn't they?

DamienCassou:
On March 30, 2018 5:02:34 AM GMT+02:00, Steve Purcell <notifications@github.com> wrote:
>Thanks! The fixes here are useful. I'm not sure that the main change
>belongs in `package-lint`, though: it's quite an edge case, and authors
>could instead add fake entries to `package-alist` for their unpublished
>packages, couldn't they?

I agree. You can still merge the other commits if you want to.
-- 
Damien Cassou
http://damiencassou.seasidehosting.st

"Success is the ability to go from one failure to another without
losing enthusiasm." --Winston Churchill

DamienCassou:
> authors could instead add fake entries to package-alist for their unpublished packages

I have tried this approach but this is not easy because package-lint takes the lowest version number in the archive (in `package-lint--lowest-installable-version-of`). So, it's not a matter of adding to the package archive, but replacing what's in there: something that package.el does not support out of the box. This starts to be quite messy in my opinion and I would appreciate some support from package-lint. After all, linters typically have configuration options.

DamienCassou:
Any news?

DamienCassou:
I believe there is a bug in package-lint that I demonstrate with the last commit. In my opinion, the test added in the last commit should pass, but it fails. I think this is because of this code in package-lint (my PR doesn't change this behavior):

```emacs
(let ((best-version (package-lint--lowest-installable-version-of package-name)))
  (when (version-list-< best-version package-version)
    ...raise error...))
```

because `best-version` contains the *lowest* version (line 1) and because package-lint raises an error if this lowest version is lower than the requested version (line 2), there will always be an error if a package appears with a low version in any archive. This happens for json-mode: it exists in version 0.1 in gnu and 1.7.0 in melpa-stable. Any request for a version greater than 0.1 will fail.

I think package-lint should get the *highest* package version from any archive and make sure the requested version is lower than that.

I can fix that in my PR if you want.

DamienCassou:
# :warning: Please merge #104 first as this PR depends on it

DamienCassou:
> I'm not sure that the main change belongs in package-lint, though: it's quite an edge case, and authors could instead add fake entries to package-alist for their unpublished packages, couldn't they?

I have tried this approach. It works but it makes the Makefile uglier than necessary: https://gitlab.petton.fr/elcouch/elcouch/commit/d9fb6b125ee817b53b515a7d745cf55f994dd982#836efb6e25a091dcb4ff8e1dbb2f0be6a5cbf14c.

DamienCassou:
ping

akirak:
I've just found this PR, and this could be a solution to #172. Is there a plan on merging this PR?

purcell:
> I've just found this PR, and this could be a solution to #172. Is there a plan on merging this PR?

I still have mixed feelings about this, since it's an ad-hoc workaround for a warning that may appear only in certain narrow circumstances, and it's now possible to make the batch `package-lint` exit with 0 even when then are warnings. I mean, taking a hard line on it, even if an author commits several package updates simultaneously, some will be built before others in MELPA, so the warning is still kind of correct. I'm not saying I want to make life inconvenient for authors in this position, but I don't think this is a huge issue given `package-lint-batch-fail-on-warnings`.

DamienCassou:
I will close this PR now. You have made it quite clear that you don't want it :-).

PR #103, opened 2018-04-16
Add support for flymake in Emacs 26+

jabranham:
Flymake in Emacs 26 is quite nice, this adds package-lint support.


purcell:
Super, thanks!

PR #104, opened 2018-04-27
Dependency too high bug

DamienCassou:
This PR refactors the "dependency too high" warning code and fixes an issue where this warning would be raised in undesired situations (see #102) for more information.

@purcell I'm making this PR so you can merge it even if you don't like the new feature in #102.


purcell:
Great, thanks - and sorry for the slow response. :-)

PR #114, opened 2018-11-11
Silence Emacs 25's byte-compiler

tarsius:
`flymake-diag-region` and `flymake-make-diagnostic` were added in
Emacs 26.  `package-lint-setup-flymake` errors out if the user tries
to use the functionality provided by this library when using an older
Emacs, but that doesn't prevent warnings when compiling this library,
which happens even on incompatible Emacsen.

Declare the functions to silence the byte-compiler.  An alternative
approach would be to distribute this library as a separate package,
but that seems a bit drastic.

There's a second commit which does unrelated cleanup.

Totally unrelated, I would recommend you move this to the `melpa` organisation, where it can sit next to `package-build`. :wink:

### DIFF
--- a/package-lint-flymake.el
+++ b/package-lint-flymake.el
@@ -1,34 +1,36 @@
 ;;; package-lint-flymake.el --- A package-lint Flymake backend  -*- lexical-binding: t; -*-
-;;
+
 ;; Copyright (C) 2018 J. Alexander Branham (alex DOT branham AT gmail DOT com)
-;;
+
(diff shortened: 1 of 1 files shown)


Fanael:
Thanks!

purcell:
The flymake library should absolutely be a separate package, since it requires a particular Emacs version.

purcell:
I merged it previously, but I'll separate it out, I think.

purcell:
Done now - heads up @jabranham, who submitted this originally. :-)

PR #115, opened 2018-12-12
Accept usage of cl-defmethod

DamienCassou:
`cl-defmethod` can extend generic definitions coming from other packages: they don't have to start with package prefix.


purcell:
Great, thanks.

PR #118, opened 2019-04-20
(provide-me) as an (provide <symbol>) alternate

rocky:


rocky:
Here is something that might be improved in this PR: when `(provide-me)` is used, the linter needs another mechanism for getting the package prefix. I used what was in the package description, although provide me in fact uses the basename of `__FILE__`.  

And perhaps it should be getting this anyway from package description rather than the provide at the end of the file.

purcell:
What is `provide-me`? It's not a standard Emacs function afaict.

rocky:
`provide-me` is part of the `load-relative` package.

 It was written so as to provide all of the benefits of provide in a package while removing the weaknesses and tedium. The computer is better suited with this kind of administrivia, rather than burdening the Emacs Lisp programmer with such chore so as she can focus on more interesting and creative matters. 

`load-relative` is a package which is part of both ELPA and MELPA and have been around for many years.

purcell:
Thanks for clarifying!

> The computer is better suited with this kind of administrivia.

Haha, I am familiar with the concept of computers. :-)

Would you mind moving the logic for determining the feature to `package-lint--provided-feature` please? It's weird factoring for the function to return `(provide-me)` as the provided feature, so this would tidy up the code.

rocky:
> Would you mind moving the logic ... 

Ok - I'll make a stab at this later when I have time.

rocky:
Looks like something changed between Emacs 24.2 and Emacs 24.3 that is causing problems. Installing 24.2 from source of via evm on Ubuntu 18.10 is not easy. If you have thoughts on what might be wrong that would save me some time. Otherwise I'll have to look into some sort of virtualization to get emacs 24.2 installed.

package-lint-test.el
```diff
@@ -421,6 +421,10 @@ Alternatively, depend on (emacs \"24.3\") or greater, in which cl-lib is bundled
     '((1 1 error "There is no (provide 'test) form."))
     (package-lint-test--run "" nil nil nil "(provide 'blargh)"))))
 
+(ert-deftest package-lint-test-error-accept-provide-me ()
```
Fanael: Drop the `error-` bit from the test name, this test doesn't check for presence of an error.

rocky: @Fanael - sure that's an easy change/fix. The harder one is understanding what is going wrong in emacs 24.2 and before. Thoughts?

package-lint.el
```diff
@@ -932,8 +932,10 @@ The returned list is of the form (SYMBOL-NAME . POSITION)."
   "Return the first-provided feature name, as a string, or nil if none."
   (save-excursion
     (goto-char (point-max))
-    (when (re-search-backward (rx "(provide '" (group (1+ (or (syntax word) (syntax symbol))))) nil t)
-      (match-string-no-properties 1))))
+    (cond ((re-search-backward "(provide-me)" nil t)
+           (file-name-sans-extension (file-name-nondirectory buffer-file-name)))
+          (t (re-search-backward (rx "(provide '" (group (1+ (or (syntax word) (syntax symbol))))) nil t)
+             (match-string-no-properties 1)))))
```
Fanael: You're calling `match-string-no-properties` even when `re-search-backward` fails here, which signals because `(0 0)` is not a valid range. Just drop the `t` and let the `re-search-backward` call be the condition.

rocky: This change seems to have fixed things - thanks! 91ee981 now has incorporated both of your excellent suggestions.

purcell: Great! My only reservation here is that it should normally be possible to call `package-lint` on a buffer that has no associated file, in which case the null `buffer-file-name` would cause `file-name-nondirectory` to explode. I don't know if there's a simple workaround, so I guess I'm not too bothered since this is in the minority `provide-me` case.

Aside from that, the code should look for `(provide 'foo)` and fall back to `(provide-me)` rather than the other way around, since the former will be by far the most common case.

rocky: > it should normally be possible to call package-lint on a buffer that has no associated file

In that situation, using `(provide-me)` in a buffer with no associated file,  makes no sense. 

I've switched the around the order of thest tests in 22a0467 .

Fanael:
Squashed and merged, thanks!

rocky:
Thanks - now I can contemplate adding more melpa packages that use this lingo.

purcell:
Super, thanks @rocky!

rocky:
@purcell over a week ago I put in a PR for such two realgud packages, e.g. https://github.com/melpa/melpa/pull/6153

I'm holding off on more PRs until I get feedback. 

I fully appreciate the work that is needed to support reviews and understand that things are often probably backlogged. But I thought I'd mention just in case.

purcell:
Great, let me just check: I didn't completely clear the backlog at the weekend, sorry.

PR #120, opened 2019-05-12
Titlecase summary

lassik:
Check that package summary starts with uppercase letter


lassik:
Fixed tests

purcell:
Thanks for this!

PR #122, opened 2019-06-06
Allow define-/defun-/list- before package prefix

lassik:
The Emacs Lisp Coding Conventions section in the GNU Emacs Lisp
Reference Manual says: "Occasionally, for a command name intended for
users to use, it is more convenient if some words come before the
package's name prefix. For example, it is our convention to have
commands that list objects named as ‘list-something’, e.g., a package
called ‘frob’ could have a command ‘list-frobs’, when its other global
symbols begin with ‘frob-’. Also, constructs that define functions,
variables, etc., work better if they start with ‘defun’ or ‘defvar’,
so put the name prefix later on in the name."

### DIFF
--- a/package-lint.el
+++ b/package-lint.el
@@ -770,7 +770,8 @@ Valid definition names are:
       (let ((prefix-re
              (rx-to-string
               `(seq string-start
-                    (or (seq ,prefix (or "-" string-end))
+                    (or (seq (opt (or "define-" "defun-" "list-"))
+                             ,prefix (or "-" string-end))
                         (seq "global-"
                              ,prefix
                              (or "-mode"
--- a/package-lint.el
+++ b/package-lint.el
@@ -770,7 +770,7 @@ Valid definition names are:
       (let ((prefix-re
              (rx-to-string
               `(seq string-start
-                    (or (seq (opt (or "define-" "defun-" "list-"))
+                    (or (seq (opt (or "define-" "defun-" "defvar-"))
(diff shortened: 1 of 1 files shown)


purcell:
Thanks, I support this in part: allowing `define-PKGNAME` / `defvar-PKGNAME` etc. is a good idea.

However, `list-` is given as an _example_ of a general principle of being allowed to choose an arbitrary prefix in the name of user convenience, and I don't want to promote that.

So I'll tweak this commit accordingly and merge it now - thanks again!

lassik:
Thanks. The quote does say, `it is our convention to have commands that list objects named as ‘list-something’`. That sounds like they prefer that form for all listing commands.

purcell:
Perhaps, but it's also only for interactive commands (which we don't explicitly look for) and they also have both `list-packages` and `package-list-packages` in `package.el`, for example. With a completing `M-x`, there's not much difference IMO.

lassik:
OK, let's leave `list-` out of the special cases. It's only a warning anyway, not an error.

lassik:
`with-` is another prefix that could be added (but I didn't find many examples of existing use). It's good to recognize common macro prefixes because Emacs auto-indents them correctly out of the box. `lisp-mode` recognizes all of `with-` and `define-` and `def`. `emacs-lisp-mode` does not recognize `with-` for some reason but it handles the others.

lassik:
LGTM 👍

PR #123, opened 2019-06-15
Check globalized-minor-mode only when package can be parsed

riscy:
(for #124)

`package-lint--check-globalized-minor-mode` binds a variable to
```elisp
(intern (package-lint--provided-feature))
```
but if the package lacks a footer this will be nil, which ultimately causes an error when the rest of the check runs.

I think this fix fits, but I'll raise an issue as well in case there's more to it than I'm seeing.


purcell:
Some of the other checks handle this in the callback invoked for each discovered symbol, like [this](https://github.com/purcell/package-lint/blob/110edb00e23eb266da068de9a404beb6202a4814/package-lint.el#L774), so it might be more consistent to do it that way instead.

purcell:
Awesome, thanks!

PR #126, opened 2019-07-07
Note if-let*, when-let* added in Emacs 26.0

amake:
These two macros were added in https://github.com/emacs-mirror/emacs/commit/be10c00d3d64d53a7f31441d42f6c5b1f75b9916, part of Emacs 26.0.


Fanael:
:+1:

PR #127, opened 2019-07-10
Output compiler-style errors and warnings in batch mode

lassik:
The output now reads like many compilers' output:

```
Entering directory '/absolute/path/to/directory'
filename:line:column: error|warning: message
filename:line:column: error|warning: message
...
```

That means Emacs' compilation-mode can make hyperlinks from the errors and lets you easily jump to the location of each error.


lassik:
@purcell @Fanael Any comment on this?

Fanael:
Sorry, I completely missed it. Looks good, thanks!

PR #128, opened 2019-07-28
Note some functions that were added in Emacs 26.1

riscy:
I think this does a good job of filling in most if not all of the functions added in Emacs 26.1.  I verified that each of these symbols are "fboundp" and that they're not duplicated in the other symbol lists.  A reference: https://www.masteringemacs.org/article/whats-new-in-emacs-26-1 (although this document mentions `url-cookie-delete-cookie` which should actually be `url-cookie-delete-cookies`).

### DIFF
--- a/package-lint.el
+++ b/package-lint.el
@@ -237,6 +237,44 @@ This is bound dynamically while the checks run.")
           dired-mouse-find-file-other-frame
           if-let*
           when-let*))
+   (cons '(26 1)
+         (package-lint--match-symbols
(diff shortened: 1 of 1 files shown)


riscy:
I believe this is out of date now -- I'll close, at least for the time being.

purcell:
Yep, this should be completely handled by #137, yay! Funny how creating `nix-emacs-ci` had the unexpected benefit of making that change possible.

PR #130, opened 2019-07-29
Enforce datetime versions for melpa packages

### LINKED ISSUE #99: Remove or downgrade "Use a non-snapshot version number for dependency"
I think this advice is not helpful, and is probably actively harmful.

I don't have much control over how my dependencies are handling releases. If they only have snapshots, or use snapshot versions, I must depend on a snapshot. So this is not actionable advice for the author of the package being checked. If we really want to discourage versioning like this, the warning should be on declaring a snapshot version, not using one.

But I think date-based versioning is probably the best option package authors have right now, as long as `version-list-<=` is the only dependency relation Emacs understands. If I am using a yet-unreleased feature I need to depend on a snapshot version, even when my dependencies are tagging releases. And with a date-based dependency the depender doesn't need to be updated again once its dependencies "go gold".

The risk of such things is that an unstable feature might "go away" but Emacs's naive version dependency resolution means I can't avoid this risk with any versioning scheme.

dickmao:
We can't and probably shouldn't change `package.el` to bend to melpa's datetime versioning.  I recommend going all-in on that versioning.

cf. melpa/melpa#2955

Closes #99

### DIFF
--- a/package-lint-test.el
+++ b/package-lint-test.el
@@ -272,14 +272,24 @@ headers and provide form."
 (ert-deftest package-lint-test-warn-snapshot-dep ()
   (should
    (equal
-    '((6 24 warning "Use a non-snapshot version number for dependency on \"package-lint\" if possible."))
-    (package-lint-test--run ";; Package-Requires: ((package-lint \"20160101.1234\"))"))))
+    '((6 24 warning "Use a snapshot version for melpa package \"package-lint\"."))
+    (package-lint-test--run ";; Package-Requires: ((package-lint \"0.2\"))"))))
+
+(ert-deftest package-lint-test-warn-non-snapshot-dep ()
+  (should
+   (member
+    '(6 24 warning "Use a non-snapshot version for non-melpa package \"org\".")
+    (package-lint-test--run ";; Package-Requires: ((org \"20190707\"))"))))
+
+(ert-deftest package-lint-test-dual-listing-dep ()
+  (should (equal '() (package-lint-test--run ";; Package-Requires: ((which-key \"20190707\"))")))
+  (should (equal '() (package-lint-test--run ";; Package-Requires: ((which-key \"3.3.0\"))"))))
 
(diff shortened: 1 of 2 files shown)


alphapapa:
No thanks.  What we have now is working pretty well, even if it isn't ideal.  Let's not double-down on relatively meaningless, monotonic timestamp versioning.

dickmao:
It's "working pretty well," because the melpa-build timestamp is not meaningless.  It is isomorphic to the git SHA1 at the time of the melpa build.  It may offend the senses, but it does have meaning and given the last four years' progress towards "the ideal," I suspect it's here to stay.

alphapapa:
That's why I said, "relatively meaningless."  For example, when I push a commit that only adds a commented line to my source file, the package is rebuilt and the version timestamp changes, even though the "executable meaning" of the package is unchanged.  Therefore the version timestamp does not necessarily indicate a change in the function of the package, and therefore it is relatively meaningless compared to a version number intentionally incremented by the author.  We should not give it additional weight or enforce it more strictly.

dickmao:
> We should not give it additional weight or enforce it more strictly.

I respect your experience but it's unclear whether you've grasped the larger point.  If it weren't for the fact that the current enforcement of semver reliably breaks dependency management, I wouldn't be doubling down on datetime either.

alphapapa:
I would love it if all Emacs package authors used SemVer-style versions, and if they all tagged stable releases for MELPA Stable.

But I don't see how enforcing timestamp-version dependency declarations helps us reach that goal.

And despite not having SemVer everywhere in MELPA, and despite the somewhat chaotic way that packages are updated by authors, MELPA seems to work well, overall.

IIUC it would also introduce some drawbacks, like having to refer to MELPA builds to get canonical timestamp-version numbers, rather than getting them directly from packages' source files or git repos.

It doesn't seem like a good solution to me.

dickmao:
Withdrawn in deference to
https://github.com/purcell/package-lint/issues/99#issuecomment-537248206

PR #136, opened 2019-09-22
Check for field numbers in format strings (Emacs 26.1+)

### LINKED ISSUE #135: Detect format field numbers
The `format` function became able to accept field numbers in placeholders [in Emacs 26.1](https://github.com/emacs-mirror/emacs/blob/a6894cf0c32c0cc50f1a45cb51117ce01446d666/etc/NEWS.26#L562):

> For example, `(format "%2$s %1$s %2$s" "X" "Y")` produces `Y X Y`.

I just found that this was tripping me up in a package of mine. It is detectable e.g. in CI if you build against pre-26.1 *and* are passing a string literal to `format` (the compiler will error out) but it is undetectable until runtime if you e.g. store the format string in a variable and pass the variable to `format`.

I would like to contribute a check for this; how should I go about it? It would be nice to be able to catch the indirect case as well, so my thought is to inspect all strings and see if they have placeholders with field numbers (under the theory that they are unlikely to be found outside of format strings).

amake:
Closes #135

### DIFF
--- a/package-lint-test.el
+++ b/package-lint-test.el
@@ -529,6 +529,10 @@ Alternatively, depend on (emacs \"24.3\") or greater, in which cl-lib is bundled
   (should
    (equal
     '()
+    (package-lint-test--run "(format \"%%1$s\" \"foo\")")))
+  (should
+   (equal
+    '()
     (package-lint-test--run ";; Package-Requires: ((emacs \"26.1\"))
 (format \"%1$s\" \"foo\")"))))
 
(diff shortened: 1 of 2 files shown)


amake:
The Travis errors seem to be connectivity problems with the GNU package archive.

purcell:
The failure is weird, and I'm taking a look at that, but the change looks good. I wonder if there's a method that users can use to escape the `%` interpolation, e.g. with `%%`, such that the regexp in this change might produce false positives?

amake:
Yes, good catch. I've fixed the escape handling.

purcell:
Really unsure why the build is failing, but I'm pretty confident it's unrelated to anything in this PR, so I'm going to merge this in any case. Thanks!

PR #137, opened 2019-09-29
Improve reporting of required emacs versions

purcell:
Okay, so using `nix-emacs-ci` and a little hackery, I have a way to fairly accurately export a full list of libraries, variables and functions available in the standard library.

This PR contains the machinery for producing that data (manually), a checked-in version of the data itself, and updates to `package-lint.el` so that the data is used at compile-time to build the lists of which functions/libraries were added in which Emacs versions.

The PR also makes `package-lint` report appropriate warnings about the use of `seq`, `let-alist` and `cl-lib` *and* their functions, indicating that the user should either depend on a newer Emacs *or* the appropriate "backport" package.

Downside is that checking is a bit slower, because the previous approach of scanning for an optimised regexp of matching functions doesn't work when there are thousands of such functions (there's a length limit for regexps, I discovered!). The memory footprint of the package is also higher.

I've already changed the MELPA recipe to include the new "data" subdir, so this would be ready to merge, but I thought I'd file a PR first for feedback. I've tested this locally, via CI, and I tried it on Flycheck and already got [useful output that wasn't available previously](https://github.com/flycheck/flycheck/pull/1603#issuecomment-536257819).

The bundled data goes back to the 24.1–24.2 changes. I might consider publishing the data itself as a separate package in due course, because it could be interesting to authors.

I discovered a surprising number of functions and libraries have been removed from Emacs over the years too: `package-lint` can easily use the data about these cases to warn about the use of any of those.

I'd be pleased to have feedback from @Fanael, @tarsius, @cpitclaudel, @riscy or anyone else who might be interested. (@riscy is probably the most active user of `package-lint`, since he's doing such an amazing job of running it on MELPA submissions!)


purcell:
I'm also not super excited about checking in a 300kb gzipped archive, which is likely to inflate the repo over time. Not sure if it's worth splitting it by Emacs version into separate files... might make sense.

purcell:
I should probably also note that the rationale for using a compressed data file was that generating the data is annoying, so I've opted to store more data than `package-lint` currently uses: I didn't want all of that loaded at runtime, but perhaps that concern is not worth the complexity cost of the code that loads it selectively at compile time...

purcell:
A further quick note about the limitations of the data, since it is not 100% accurate: it is produced by loading every possible stdlib library inside a clean Emacs, then exporting all the known features, vars and functions/macros.

However, the Emacs versions used are not built with lots of optional features. Some libraries will fail to load because their C functions are unavailable: `dbus` would be an example, and its symbols will not show up in the data.

Only a very small minority of symbols will be affected by this issue, and given that we are mainly using _differences_ between the symbol sets for consecutive Emacs versions, I don't feel like this is a big deal.

tools/sym-diff.el
```diff
@@ -0,0 +1,39 @@
+;;; sym-diff.el --- Insert description here -*- lexical-binding: t -*-
```
Fanael: Probably needs copyright information, at least SPDX header at the very minimum to be safe.

(I know I screwed that up three years ago when I added the test runner.)

tools/sym-dump-all
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env bash
```
Fanael: Ditto.

tools/sym-dump.el
```diff
@@ -0,0 +1,58 @@
+;;; sym-dump.el --- Insert description here -*- lexical-binding: t -*-
```
Fanael: Ditto.

package-lint-test.el
```diff
@@ -333,8 +333,8 @@ Alternatively, depend on (emacs \"24.3\") or greater, in which cl-lib is bundled
    (equal
     '()
     (package-lint-test--run
-     ";; Package-Requires: ((emacs \"25\"))
-\(when-let ((foo (bar))) (message \"ok\"))"))))
+     ";; Package-Requires: ((emacs \"25.1\"))
+(when-let ((foo (bar))) (message \"ok\"))"))))
```
Fanael: I remember this – unescaped `(` at the start of a line – causing some navigation or parsing problems in the past, but I don't remember what was the issue nor how to reproduce it anymore.

Fanael:
That looks very impressive, nice work, I never even though about approaching this problem in this way!

> Downside is that checking is a bit slower, because the previous approach of scanning for an optimised regexp of matching functions doesn't work when there are thousands of such functions (there's a length limit for regexps, I discovered!). The memory footprint of the package is also higher.

Is it noticeable at all? I can try this improved detection on some older hardware (I despise the notion that everyone has the latest hardware that silicon valley techbros are so fond of assuming) to see if it's still usable there, and if it is, I wouldn't worry about it at all; if it is not, we'll think about optimization then.

> I'm also not super excited about checking in a 100kb gzipped archive, which is likely to inflate the repo over time. Not sure if it's worth splitting it by Emacs version into separate files... might make sense.

Probably safer to do it that way if you're worried about bloating the repository too much, OTOH keeping it in one file is just simpler and with how often new emacs versions are being released that would just bloat the repo by just a single new ~100 KB blob every new release, no?

> I should probably also note that the rationale for using a compressed data file was that generating the data is annoying, so I've opted to store more data than `package-lint` currently uses: I didn't want all of that loaded at runtime, but perhaps that concern is not worth the complexity cost of the code that loads it selectively at compile time...

We may start using that extra data in the future if we add some new feature, so it's fine.

package-lint-test.el
tarsius: It messes with `outline-minor-mode`, which is led to believe that this line begins a top-level expression.

package-lint.el
```diff
@@ -61,187 +62,70 @@
...
+                                             (if load-file-name
+                                                 (file-name-directory load-file-name)
+                                               default-directory)))
+                          (read (current-buffer)))))
+
+    (defconst package-lint--libraries-added-alist
+      (mapcar (lambda (version-data)
+                (let ((version (car version-data))
+                      (added-functions (let-alist (cdr version-data) .features.added)))
+                  (cons version (funcall 'package-lint--match-symbols added-functions))))
```
Fanael: What's the reason for `funcall` there?

cpitclaudel:
Great work!  I wouldn't worry about 100kB, since we don't expect it to keep growing indefinitely (I don't imagine you'll want to check backwards compatibility across more than 10 releases at most)

Fanael:
> I don't imagine you'll want to check backwards compatibility across more than 10 releases at most

The period from 24.1 to 26.3 is 11 releases already.

package-lint.el
purcell: Good catch, no good reason at this point, will fix. :-)

package-lint-test.el
purcell: I think I'm not going to worry too much about that at this point.

purcell:
Thanks for the feedback, all! I've made a few little changes as a result.

Alright, I'll merge this and we'll see if anything explodes. :-)

riscy:
Just chiming in to say this is awesome.  A 100kb add doesn't seem outrageous to me at all!

purcell:
Hah, discovered a few little issues, so scrambled to fix them. Dumping wasn't working great with some Emacs versions, it seems.

DamienCassou:
Great work!

tarsius:
[So it has begun](https://github.com/melpa/melpa/issues/4154).

purcell:
Haha, yes, the effort continues...

PR #140, opened 2019-10-09
straight.el support (fix #108)

### LINKED ISSUE #108: `straight.el` is not supported
Exclusively using [`straight.el`](https://github.com/raxod502/straight.el) to install packages is not supported :`package-lint--check-packages-installable` says "Package … is not instalable" even though I have the package installed using `(straight-use-package …)`

FelipeLema:


package-lint-test.el
```diff
@@ -39,6 +39,8 @@ VERSION is a list of numbers, e.g., (0 5 0) to represent version
                                       (:url . "https://gitlab.com/somewhere"))])
    (or archive "melpa-stable")))
 
+
+
```
tarsius: Looks like you are accidentally adding whitespace.

package-lint-test.el
```diff
@@ -270,6 +272,26 @@ headers and provide form."
     '((6 23 error "Package example-nonexistent-package is not installable."))
     (package-lint-test--run ";; Package-Requires: ((example-nonexistent-package \"1\"))"))))
 
+(defvar straight--recipe-cache) ;; simulate straight.el internal
+
+(ert-deftest package-lint-test-warning-straight.el-installed-dep ()
+  (setq straight--recipe-cache (or
+				(and (boundp 'straight--recipe-cache)
```
tarsius: Indention seems wrong. You might want to double check this in Emacs with `whitespace-mode` turned on.

tarsius:
You should squash this into a single commit (we are not interested in knowing that you initially forgot to "declare straight.el internal variable" for example) and explain *what* this does (beyond just mentioning #108) and also *why*. That explanation should probably go into the commit message and also be posted here as the pull-request description.

purcell:
FWIW, I'm not keen to merge any change which makes explicit references to `straight.el`. That's just not good factoring of the code: what happens when another package like `straight` comes along.

Rather, I would _consider_ providing a customizable function which `package-lint` would use to obtain a list of installable packages, such that the default implementation would be to use the `package.el` list.

But note that I'm still not even very convinced that this is a genuinely useful use case, as I noted in my comments on #108, and the motivation for this has not really been described there or here.

tarsius:
> what happens when another package like `straight` comes along?

Like my `borg` for example. :wink: 

Actually I was thinking about this earlier today, *before* I saw this. I haven't actually studied the existing code or this patch, but my intermediate understanding was that explicit support for other package managers is not necessary. Isn't it enough to add Melpa to `package-archives` and call `package-refresh-contents` occasionally? Even if one never installs any packages using `package.el`, that should be enough to keep `package-lint` happy and the only drawback is that `~/.emacs.d/elpa/` will be created, right?

Even as the author of an alternative package manager I think it is more important to ensure that the dependencies are available when using `package.el`--after all that's what most people use.

purcell:
> Isn't it enough to add Melpa to `package-archives` and call `package-refresh-contents` occasionally? Even if one never installs any packages using `package.el`, that should be enough to keep `package-lint` happy and the only drawback is that `~/.emacs.d/elpa/` will be created, right?

Yeah, exactly, and `package-user-dir` could be overridden to be a temp dir if necessary.

> Even as the author of an alternative package manager I think it is more important to ensure that the dependencies are available when using `package.el`--after all that's what most people use.

That's exactly the idea.

But still, I'll wait to hear more from @FelipeLema before closing this, even though I feel that's the likely outcome.

FelipeLema:
@purcell I hadn't understood your intention to keep `package-lint` handling packages by using default provided features from Emacs. Some packages provide optional support to external packages, althought `package-lint` is not going to be the case.

Now that I do, I agree that a better way to handle alternative package managers is for `package.el` to provide extension points.

Personally, I like the "variable that holds a list of functions that may claim authority on feature" + "`cl-defmethod` for a given feature, fallback to `cl-defgeneric`" model (like `project.el` and `xref`). 

> Isn't it enough to add Melpa to package-archives and call package-refresh-contents occasionally?

That's one way to make `package-lint` work, although instead now I have to keep two package installers instead of one, which misses the point of why some of us use `straight.el` in the first place.

As reference, the original problem was that `package-lint` would complain about a package being detected as "not installed" and stop checking the rest of the file (using `flymake`). This hindered my package development, because other issues in the rest of the file would be reported by other users when registering the code as a new package in MELPA. Maybe there's a alternative way to deal with this problem that I'm not considering (like using a separate "build" configuration with CI).

Thanks for taking the time to ~~listen~~ read.

tarsius:
I am starting to add something like this to my `Makefile`s:

```elisp
lint: $(ELS)
	@$(EMACS) -Q --batch $(EMACS_ARGS) $(LOAD_PATH) \
	  -L ../package-lint -l package-lint \
	  --eval "(add-to-list ' \
	  '(\"melpa\" \"https://melpa.org/packages/\") t)" \
	  --eval "(add-to-list 'package-archives \
	  '(\"org\" \"https://orgmode.org/elpa/\") t)" \
	  -f package-initialize \
	  -f package-lint-batch-and-exit $<
```

@FelipeLema This might or might not work for you. (You still have to occasionally run `package-refresh-contents` in an interactive session.)

@purcell Any suggestions on how to make that nicer? Maybe `package-lint-bratch-and-exit` could be taught to add the additional `package-archives` and call `package-initialize` itself?

FelipeLema:
@tarsius I was looking into something like that (calling emacs to lint my code using a separate command).
I will close this PR since I understand it's not going to be accepted, although I will provide a separate PR with an extension point so other people (me) using non-default package managers (straight.el) can handle package dependencies.

PR #143, opened 2019-10-17
Warn if indentation is inconsistent with elisp defaults

purcell:
I suspect this one's a bit slow. What do you think, @Fanael?

### DIFF
--- a/package-lint-test.el
+++ b/package-lint-test.el
@@ -116,7 +116,7 @@ headers and provide form."
     (should-not (package-lint-test--run "(defcustom test-something (kbd \"C-g\"))"))
 
     ;; Don't bind a key sequence ending in <ESC> except following another <ESC>
-    (should (equal (package-lint-test--run "(defcustom test-something (kbd \"C-x <ESC>\")")
+    (should (equal (package-lint-test--run "(defcustom test-something (kbd \"C-x <ESC>\"))")
                    `((6 43 warning ,reserved-message))))
     (should-not (package-lint-test--run "(defcustom test-something (kbd \"C-x <ESC> <ESC>\"))"))))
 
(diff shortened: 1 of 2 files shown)


purcell:
Thinking about it, there's probably a way to exit the check early when the first incorrectly-indented line is detected, but in the common case of everything being fine, this is still going to be a costly check.

purcell:
An interesting aspect of this is that the test relating `with-eval-after-load` actually produces an additional "incorrect indentation" error when run in Emacs versions prior to the addition of that macro, because they don't know to indent it correctly. This would be an example of code linting differently between Emacs versions, so while it would be easy enough to adjust the test accordingly, it might already be a sign that this new check is undesirably.

Fanael:
> I suspect this one's a bit slow. What do you think, @Fanael?

I wouldn't worry about speed here. Evaluating `(benchmark-run-compiled 10 (package-lint-buffer))` in recent `helm.el` on my Haswell box (i5-4590, so not great by today's standards, but still likely better than many laptops; if you want, I can benchmark it on something like a Pentium III or a VIA Eden :stuck_out_tongue:):
- master: `(4.549648831 9 0.307750337)`
- this PR: `(5.314536742 14 0.4848494670000001)`

Only 9% of time is spent in `package-lint--check-consistent-indentation`; `package-lint--check-macros-functions-available-in-emacs` remains an order of magnitude hotter.

> An interesting aspect of this is that the test relating `with-eval-after-load` actually produces an additional "incorrect indentation" error when run in Emacs versions prior to the addition of that macro, because they don't know to indent it correctly. This would be an example of code linting differently between Emacs versions, so while it would be easy enough to adjust the test accordingly, it might already be a sign that this new check is undesirably.

This is much more damning IMO, this patch is very brittle, not only for us when maintaining tests, but also for users, as it heavily relies on Emacs having accurate indentation info of each function and macro. Do we want to break linting under `emacs -Q` without having to load the lintee and its dependencies first? Isn't this how flymake and flycheck operate?

I personally consider that a reason to NAK it.

purcell:
Yep, I think I agree. I've been picking out a couple of things that come up in MELPA reviews, and (like the unidiomatic closing parens on their own lines) this issue was one. I guess ultimately it's an additional check we could easily just incorporate there.

/cc @riscy

conao3:
I +1 this PR. It is useful for my developing packages and checking other packages code.

purcell:
> I +1 this PR. It is useful for my developing packages and checking other packages code.

Yep, but I think we can't merge this in its current form simply because the indentation will give different results in different Emacs versions and with different packages loaded.

conao3:
hmm, I understand the big issue related to this PR.
To solve that, I suggest installing dependency packages via Cask and linting indentation another clean Emacs process.
But as this package cannot depend on Cask, create another package is a simple way to achieve goal.

purcell:
Yes, but that's not a solution, because even the Emacs version will make a difference: if one Emacs recognises, say, `when-let` as a built-in macro with a certain indentation scheme, it will indent the code differently than a different Emacs version that doesn't know about `when-let`. If we can't all get reproducible lint results even when using different Emacs versions, that problematic IMO.

conao3:
Anyway, I'm going to implement that separated lint package.
The consent here is that merging this feature into package-lint is likely to be difficult.

purcell:
Additionally, `package-lint` works entirely without needing other elisp code to be loaded. This is important for usage in MELPA, where we can only integrate it in our build process if we can be sure that no potentially-unsafe elisp code will be evaluated by our build server. Requiring dependency packages to be installed and loaded (in order for Emacs to understand the indentation rules for macros they define) would be very bad in that environment.

purcell:
Closing due to issues described previously.

PR #146, opened 2019-11-03
allow `with` prefix

conao3:
CC: @lassik

Hi!
I add allow `with` prefix rule.
This idea come from this [comment](https://github.com/purcell/package-lint/pull/122#issuecomment-499829385)

### DIFF
--- a/package-lint.el
+++ b/package-lint.el
@@ -735,7 +735,7 @@ Valid definition names are:
 - a NAME whose POSITION in the buffer denotes a global definition."
   (or (string-prefix-p prefix name)
       (string-match-p package-lint--sane-prefixes name)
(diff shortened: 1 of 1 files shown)


Fanael:
:+1:

conao3:
Thanks!

lassik:
LGTM 👍

purcell:
This is fine, but I'm generally a little -1 on allowing these various prefixes, because there are usually reasonable possible alternatives, and these prefixes aren't often used anyway: I wouldn't want to actively encourage them.

lassik:
`emacs-lisp-mode` doesn't have a special indentation rule for the `with-` prefix, in Common Lisp `with-` is popular, so arguments could be made either way.

Fanael:
@purcell Of course, but `with-` macros are a longstanding lisp idiom, which is why I made an exception.

PR #150, opened 2019-11-22
Warn when package summary end with period

conao3:
Hi!

Proposal that summary should not end with period.
Currently, almost packages' summary doesn't end with period but some packages not.


purcell:
Makes sense, thanks! Would you mind changing "should not end with period" to "should not end with a period" please?

conao3:
OK!

purcell:
Great, thank you!

conao3:
Thanks!

PR #151, opened 2019-11-23
Accept just quoted list

conao3:
I found this issue, but I cannot solve this issue.
I add failed test, so please solve this testcase.

``` emacs-lisp
(defvar test-some-list '(
                         ;; some comment
                         (defun defvar defcustom)))
```

This code is clean, but this package warn prefix issue.

### DIFF
--- a/package-lint-test.el
+++ b/package-lint-test.el
@@ -425,6 +425,12 @@ Alternatively, depend on (emacs \"24.3\") or greater, in which cl-lib is bundled
   (should (equal '() (package-lint-test--run
                       "(defun org-dblock-write:test ())"))))
 
+(ert-deftest package-lint-test-accept-just-quoted-list ()
+  (should (equal '() (package-lint-test--run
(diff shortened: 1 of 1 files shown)


purcell:
Yep, it would be nice to handle this more robustly, though note that due to the very dynamic nature of Lisp there will always be cases that `package-lint` cannot understand clearly without evaluating the whole file. In the meantime, you can also use `(list (list 'defun 'defvar 'defcustom))` as a workaround. :-/

conao3:
OK, thanks! I sent this PR based Test-driven development, but it isn't nice to keep this PR long time in PR list.
I will change this report to issue.

purcell:
> OK, thanks! I sent this PR based on the idea of Test-driven development, but it isn't nice to keep this PR long time in PR list.
> I will change this report to issue.

Feels like this just makes more work for everyone. It's convenient having the commit with the test case.

conao3:
Sorry, long time no response. I had not able to understand your intended...
I would reopen this PR?

PR #159, opened 2020-01-10
Use local elpa directory

conao3:
Hi!
I found this package `run-tests.el` may change global user `.emacs.d` directory.
As this improvement, the Emacs via `run-tests.el` use local `elpa` directory separated from user global `.emacs.d`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 *.elc
 /.symbol-dumps
+/elpa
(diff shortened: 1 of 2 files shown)


conao3:
Oops, this PR triggers GitHub Actions too. Full CI log is [here](https://github.com/conao3/package-lint/commit/74f9de0c02e36208331c9c4addd046aa73d6f3bd/checks?check_suite_id=394453525) and had passed all checks.

purcell:
I'm okay with the side effects here, so prefer not to merge this. I might consider a more complete sandboxing solution in future though.

conao3:
OK, thanks!

PR #162, opened 2020-02-23
Remove the "summary should start with an uppercase letter" check

stepnem:
Such a requirement isn't documented in (info "(elisp)Coding Conventions")
or (info "(elisp)Library Headers"), checkdoc doesn't require it, many
libraries out there (including Emacs core and MELPA ones) don't follow
it, and often it just doesn't make sense, e.g. when the first word is
a command name, cf.

  https://github.com/melpa/melpa/pull/6354#issuecomment-520126497

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ Roadmap
 
 Additional checks for future versions:
 
-- WARN: header line formatting / capitalisation
+- WARN: header line formatting
 - trailing line presence / formatting
 - WARN: library is called *-mode but doesn't provide a major mode
 - checkdoc failures for interactive functions / defcustoms
--- a/package-lint.el
+++ b/package-lint.el
@@ -672,11 +672,6 @@ DESC is a struct as returned by `package-buffer-info'."
        'warning
        "Package should have a non-empty summary."))
      (t
-      (unless (let ((case-fold-search nil))
-                (string-match-p "^[A-Z0-9]" summary))
-        (package-lint--error-at-bob
-         'warning
-         "The package summary should start with an uppercase letter or a digit."))
       (when (> (length summary) 60)
         (package-lint--error-at-bob
          'warning


purcell:
Thanks for taking the time to submit this.

> Such a requirement isn't documented in (info "(elisp)Coding Conventions")
> or (info "(elisp)Library Headers"),

This applies to several things we detect.

> checkdoc doesn't require it, many
> libraries out there (including Emacs core and MELPA ones) don't follow
> it,

Many existing libraries contain stylistic or packaging mistakes.

> and often it just doesn't make sense, e.g. when the first word is
> a command name, cf.

It's always possible to write a good short description that follows this suggestion, and the result is more readable package descriptions.

For example, in place of "vcsh integration" I might write "Configure and query vcsh repositories", which is more informative.

So, you're *right* that this doesn't need to be a hard rule, but it was introduced because it *generally* leads to better package descriptions, and I prefer to retain it.

stepnem:
> Thanks for taking the time to submit this.

Thank you for your consideration.

> Many existing libraries contain stylistic or packaging mistakes.

But this is not one of them, right? If you do consider this a mistake,
please submit a bug report to Emacs upstream.

> It's always possible to write a good short description that follows
> this suggestion, and the result is more readable package descriptions.

More readable in some people's _opinion_.

> So, you're *right* that this doesn't need to be a hard rule, but it
> was introduced because it *generally* leads to better package
> descriptions, and I prefer to retain it.

The problem I have with this attitude is that tools like package-lint
are often used in automatized workflows, and having to work around
cases like this kind of defeats their purpose.

Many of us have all kinds of opinions about best practices, but I
strongly believe that tools like package-lint should serve the
community by sticking to those that are actually documented and
generally accepted, instead of trying to push their maintainers'
personal opinions or agendas, however passionate or justified they
might feel about them.

purcell:
Yes, there's definitely some degree of opinion going on here. On the other hand, after many thousands of package submissions, you're the first person to dispute this specific point. :-)

I believe a lot of details like this which may seem frivolous to you have contributed to MELPA's growth and value. We've encountered issues that simply didn't exist in the past, because there were never thousands of installable packages previously: rules like requiring the symbol prefix in .el files to match the containing package name are a result of this, and are not necessarily enshrined in the elisp conventions (though on that particular point, Stefan Monnier has been supportive).

But I'm sympathetic, and I think the solution here is different levels of warnings, tbh. Currently the real issue is that `warning`s from `package-lint` are considered equally "fatal" to linting as "errors". We wouldn't decline a package just because of capitalisation in the description line, but conflating the two levels indeed makes an automated workflow less practical. I think there's an open issue about that.

stepnem:
> you're the first person to dispute this specific point. :-)

Most people don't care, apparently including MELPA reviewers (the
relative amount of packages in MELPA following this particular
requirement doesn't seem to be any different from that outside MELPA,
which is another reason for its removal: it's just noise without any
consequence other than resource waste).

"It's usually easier to ignore or work around something locally than
to try to reach consensus or change things upstream." seems well
established in Emacs community.

> I believe a lot of details like this which may seem frivolous to you
> have contributed to MELPA's growth and value.

Maybe, but let's not vaguely generalize.

> We've encountered issues that simply didn't exist in the past,
> because there were never thousands of installable packages
> previously: rules like requiring the symbol prefix in .el files to
> match the containing package name are a result of this, and are not
> necessarily enshrined in the elisp conventions (though on that
> particular point, Stefan Monnier has been supportive).

Symbol prefixes (and their consistency with package names) are solving
real, practical problems. The summary capitalization requirement, if
anything, _creates_ them.

> [...]
> 
> Currently the real issue is that `warning`s from `package-lint` are
> considered equally "fatal" to linting as "errors".
>
> [...]
> 
> I think there's an open issue about that.

Not that I can see looking at the open issue titles...

purcell:
> On 24 Feb 2020, at 10:03, stepnem <notifications@github.com> wrote:
> 
> I think there's an open issue about that.
> 
> Not that I can see looking at the open issue titles...
> 


True. I’ve just committed a fix for that - thanks for the prod.

-Steve

stepnem:
> True. I’ve just committed a fix for that - thanks for the prod.

Thanks! There seems to have been a doc string glitch, please see #164.

lassik:
@stepnem The capitalization check is my fault (PR #120). The reason is that the listing at melpa.org or in `M-x list-packages` is easier to read with consistent capitalization across packages. Whether capitalization is required or not, is not my call to make.

> "It's usually easier to ignore or work around something locally than to try to reach consensus or change things upstream." seems well established in Emacs community.

While that's probably true, it's one of the main stumbling blocks of the Emacs community. To make something easy to use, there need to be community-wide standards. Whether they are _de facto_ or _de jure_ standards doesn't really matter, as long as they are generally followed. Having _n_ slightly different ways to do the same thing is confusing to users and duplicates effort among programmers. To that effect, things like coding conventions and elaborate linter warnings are a good thing.

stepnem:
> @stepnem The capitalization check is my fault (PR #120). The reason is
> that the listing at melpa.org or in `M-x list-packages` is easier to
> read with consistent capitalization across packages.

I think I have explained my arguments why the check shouldn't be there
already.

I am very fastidious about consistency and the code I write, have been
following the conventions (the _actual_ conventions) for more than 10
years and will continue to do so, MELPA or package-lint notwithstanding.
Thus this PR and the related discussion.

Once again: there is no convention to capitalize summaries, and with
this check package-lint goes against its own purpose as stated in its
README. Please check some representative sample of Emacs libraries, both
in and outside MELPA and see for yourself.

If for some reason you or somebody else can't stand lowercase summaries,
you could "fix" that on your or MELPA side.

Personally (i.e., none of this should really matter, unlike actual
arguments) I respect your opinion of course, but don't understand it:
the summaries are not sentences, there is no reason for them to start
with a capital, similarly to e.g. command or option summaries when you
do 'ls --help' or whatever. They also don't end with a period. But
that's all just *opinions*, what matters is that package-lint IMNSHO
should only check actual conventions or other "issues that may make your
package uninstallable or unusable for some users" (again quoting the
README).

> Whether they are _de facto_ or _de jure_ standards doesn't really
> matter, as long as they are generally followed.

That doesn't make any sense. "generally followed" is what makes
something "de facto", as opposed to "de iure". (And summary
capitalization is neither de facto nor de iure, unless you want to make
your own "conventions", in which case please rename package-lint to
melpa-lint or whatever.)

> Having _n_ slightly different ways to do the same thing is confusing
> to users and duplicates effort among programmers. To that effect,
> things like coding conventions and elaborate linter warnings are a
> good thing.

Indeed, that's why this should be removed. :-)

Anyway, please let's not waste our time here. I have stated the facts
and arguments; as long as the response is "but this is better" or
"easier to read", I don't have anything more to add.

I have said this already, too, but: if you _do_ care about _actual_
conventions and believe that summary capitalization should be one of
them, please submit your request to Emacs upstream so that it is
included in the documentation.

lassik:
I just thought it was a trivial way to improve the listing. I had no idea people have strong feelings about it :) That's fine; I appreciate that you voice them.

> "generally followed" is what makes something "de facto", as opposed to "de iure".

A _de jure_ standard can also be generally followed (e.g. most laws). In software the distinction usually doesn't matter as much, as long as the end result works well.

> in which case please rename package-lint to melpa-lint

You may have a point.

What's the real issue here? Obviously nobody feels as strongly as we do about capitalization. Is this more about MELPA vs ELPA cooperation, or where the decision-making power should be?

lassik:
As for the capitalization itself, my argument is not that capitalized or not-capitalized is easier to read or more proper; my argument is that we should pick one choice and use it consistently. Just like tabs vs spaces, you can argue either side but the main point is to be consistent.

stepnem:
> I just thought it was a trivial way to improve the listing. I had no
> idea people have strong feelings about it :) That's fine; I appreciate
> that you voice them.

Anyone's feelings are not the point here (and, as you seem to contradict
yourself further in your reply, you apparently do have strong feelings
about this particular point).

The main _practical_ issue with package-lint doing this check (and that
was also discussed upthread already) is that in an automated workflow
that expects only "success" or "failure", this leads to a "failure",
same as, say wrong requires or other serious issues.

lassik:
> as you seem to contradict yourself further in your reply, you apparently do have strong feelings about this particular point

I was just trying to be polite. My communication style is not as literal as that of many hackers.

I started a new issue about the problem of message levels more generally: #169 You may have a good perspective on that issue. This capitalization thing is a small sub-issue of it.

PR #170, opened 2020-03-21
Provide some support for packaged custom themes

milouse:
This PR adds the following features:

- Do not warn user about a missing `(provide 'feature)` when providing a custom theme;
- Instead warn user about a missing `(provide-theme 'feature)` when providing a custom theme and no such line appears in the current file;
- Check that the current file ends with `-theme` when a `(deftheme` appears in the current file.

It's the first time I participate for this project, thus do not hesitate to blame me if I did anything wrong (code guideline…).

Have a good evening,

### DIFF
--- a/package-lint.el
+++ b/package-lint.el
@@ -161,6 +161,9 @@ published in ELPA for use by older Emacsen.")
             (package-lint--check-objects-by-regexp
              "(define-global\\(?:ized\\)?-minor-mode\\s-"
              #'package-lint--check-globalized-minor-mode)
+            (package-lint--check-objects-by-regexp
+             "(deftheme\\s-"
+             #'package-lint--check-deftheme)
             (when prefix
               (package-lint--check-objects-by-regexp
                (concat "(" (regexp-opt '("defalias" "defvaralias")) "\\s-")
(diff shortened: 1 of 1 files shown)


purcell:
Thanks for this! A few extra tests would be helpful.

> * Do not warn user about a missing `(provide 'feature)` when providing a custom theme;
> * Instead warn user about a missing `(provide-theme 'feature)` when providing a custom theme and no such line appears in the current file;
> * Check that the current file ends with `-theme` when a `(deftheme` appears in the current file.

I like the thinking here, but I'm not sure these tests can necessarily be done this strictly: [here](https://github.com/purcell/color-theme-sanityinc-tomorrow/blob/master/sanityinc-tomorrow-blue-theme.el) is a theme file which legitimately does not contain a `deftheme` form (though its macroexpansion does) and [here](https://github.com/purcell/color-theme-sanityinc-tomorrow/blob/master/color-theme-sanityinc-tomorrow.el) is a file which contains a `deftheme` form but legitimately is not called `*-theme.el` *and* contains a valid `provide` form.

milouse:
Yes, I was thinking about this kind of situation while doing my work, but I didn't know any live example. Thanks for pointing me this, I'll see if I can do something about it.

purcell:
Going to close this for now.

PR #174, opened 2020-04-15
Use string= to compare filenames

riscy:
I was running into some issues until I discovered the call to `eq`.


purcell:
Oh, weird, they're usually pretty interchangeable. But yes, better to be explicit - thanks!

Fanael:
They're not interchangeable on strings, two different string objects (object identity being what `eq` compares) can consist of the same characters (which is what `string=` compares).

`eq` does the right thing pretty much only for fixnums and symbols, and with the introduction of bignums in emacs 27 even using it for fixnums is iffy, because integers aren't necessarily fixnums anymore.

purcell:
Makes sense -- I was probably thinking of `equal`.

PR #176, opened 2020-06-09
Warn about requiring cl-lib <=0.5 on Emacs >=24.3

lassik:
Ping @riscy

### DIFF
--- a/package-lint-test.el
+++ b/package-lint-test.el
@@ -327,6 +327,12 @@ headers and provide form."
 Alternatively, depend on (emacs \"24.3\") or greater, in which cl-lib is bundled.")
     (package-lint-test--run ";; Package-Requires: ((cl-lib \"1\"))"))))
 
+(ert-deftest package-lint-test-warning-cl-lib-0.5-not-needed ()
+  (should
+   (member
+    '(6 52 warning "An explicit dependency on cl-lib <= 0.5 is not needed on Emacs >= 24.3.")
+    (package-lint-test--run ";; Package-Requires: ((emacs \"24.3\") (cl-lib \"0.5\"))"))))
+
 (ert-deftest package-lint-test-warning-cl ()
   (should
    (member
--- a/package-lint.el
+++ b/package-lint.el
@@ -374,7 +374,7 @@ Return a list of well-formed dependencies, same as
               (package-lint--check-packages-installable deps)
               (package-lint--check-deps-use-non-snapshot-version deps)
               (package-lint--check-deps-do-not-use-zero-versions deps)
-              (package-lint--check-do-not-depend-on-cl-lib-1.0 deps)
+              (package-lint--check-cl-lib-version deps)
               deps))
         (error
          (package-lint--error-at-bol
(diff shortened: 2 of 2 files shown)


purcell:
Thanks, I think this makes sense... would you be able to add a test case please?

purcell:
Awesome, thank you!

lassik:
Is 0.5 the right cl-lib version btw?

purcell:
Good question... it's probably `< 1.0` we need to worry about, actually. The version included in Emacs 24.3 is 1.0.

purcell:
^ I've tweaked this accordingly in 8edc96b

purcell:
Interestingly, the cl-lib version bundled in Emacs is always 1.0, so even if `cl-lib 0.7` has additions relative to the `cl-lib 1.0` bundled in Emacs 24.3, there's no way to declare a dependency on it.

lassik:
Since `cl-lib 1.0` satisfies all dependency declarations of the form `cl-lib 0.x`, isn't it impossible to meaningfully declare a dependency on `cl-lib` in Emacs >= 24.3?

conao3:
On the other hand, all packages should enable `lexical-binding`. [ref comment](https://github.com/cask/cask/issues/480#issuecomment-641140359).

This requirement requires a modern package (and that it work with most Emacs packages), it is reasonable to set Emacs-24.4 as the minimum requirement for Emacs is.
From this requirement, it may be possible to simplify the check, I think.

lassik:
Wasn't the `lexical-binding` declaration introduced in Emacs 24.1?

purcell:
> Since `cl-lib 1.0` satisfies all dependency declarations of the form `cl-lib 0.x`, isn't it impossible to meaningfully declare a dependency on `cl-lib` in Emacs >= 24.3?

Yes, but I guess I'm allowing for a world in which ELPA might eventually get cl-lib 1.x.  Or perhaps I'm misunderstanding your question?

lassik:
However, since cl-lib was added in 24.4, perhaps 24.4 would be a useful minimum version. In practice, it's probably not possible to use a large number of MELPA packages (i.e. a large number of the useful Emacs packages that exist anywhere) without cl-lib.

purcell:
`cl-lib` was added in 24.3.

lassik:
You can install cl-lib on Emacs <24.4, but it might be easier just to upgrade :) 24.4 was released almost six years ago.

lassik:
Ah, right :D Well, that's more than 7 years ago.

lassik:
https://www.gnu.org/software/emacs/history.html

purcell:
> You can install cl-lib on Emacs <24.4, but it might be easier just to upgrade :) 24.4 was released almost six years ago.

Yes, true, but `package-lint` isn't here to tell people when to start using newer versions, just to help them understand what technically works. In practice, I've written a couple of packages that are fine with just `(emacs "24.1") (cl-lib "0.5")` deps.

conao3:
I miss memorizing the dependency...
Sorry for the noise. After all, the machine should do these checks :)

lassik:
> > Since `cl-lib 1.0` satisfies all dependency declarations of the form `cl-lib 0.x`, isn't it impossible to meaningfully declare a dependency on `cl-lib` in Emacs >= 24.3?
> 
> Yes, but I guess I'm allowing for a world in which ELPA might eventually get cl-lib 1.x. Or perhaps I'm misunderstanding your question?

If you mean manually specifying cl-lib >= 1.1 in the deps, that makes sense.

Would it be logical for package-lint to warn whenever cl-lib <= 1.0 is mentioned in deps?

lassik:
Slightly off-topic, but Repology is an awesome site for checking out which distros ship with what Emacs version btw: https://repology.org/project/emacs/versions

PR #178, opened 2020-06-10
Fix package-lint's warnings about itself

lassik:
196:65: error: You should depend on (emacs "24.4") if you need
`package-desc-from-define'.

10:72: warning: An explicit dependency on cl-lib <= 1.0 is not needed
on Emacs >= 24.3.

### DIFF
--- a/package-lint.el
+++ b/package-lint.el
@@ -7,7 +7,7 @@
 ;; URL: https://github.com/purcell/package-lint
 ;; Keywords: lisp
 ;; Version: 0
-;; Package-Requires: ((cl-lib "0.5") (emacs "24.1") (let-alist "1.0.6"))
+;; Package-Requires: ((emacs "24.4") (let-alist "1.0.6"))
 
 ;; This program is free software; you can redistribute it and/or modify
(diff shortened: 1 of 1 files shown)


conao3:
Drop CI for Emacs-{24.1~24.3} is also reasonable?
https://github.com/purcell/package-lint/blob/8edc96b5086e518be5a2feb2c705c39ddc28ff6e/.github/workflows/test.yml#L16-L18

lassik:
We just discussed in #176 that package-lint needs to be able to lint packages meant for at least Emacs 24.1 and up.

Does that mean package-lint itself must also be able to run in those versions of Emacs?

purcell:
@conao3 Yep, I understood the context of your suggestion. Just wanted to clarify. :-)

lassik:
> Yes, but this thread intends to change require minimum Emacs

The purpose of this PR was not  to change the minimum Emacs version on which package-lint runs, just to ensure that `package-desc-from-define` is available. Dropping support for Emacs <24.4 was the easiest way to accomplish that, but @Fanael added a shim which does the same job while keeping those Emacs versions.

PR #179, opened 2020-06-11
Suggest Emacs-24.1 instead of Emacs-24 when enable lexical-binding

conao3:
Suggest Emacs-24.1 instead of Emacs-24 when enable lexical-binding.

Most warnings are recommended in version format with one decimal
place.  In fact, it was introduced somewhere in the development
version of 24.0.x, so specifying 24.1 is more reasonable than
specifying 24.


purcell:
Thanks

conao3:
Thanks!

Issue #181, opened 2020-08-13
Add a "when was this added?" command


tarsius:
Please add a query command that informs the user when a certain interactively read function/variable was added.

I am aware this package is about linting but it has all the necessary data and it would complicate matters if that had to be duplicated in another package just because this command might not fix perfectly here. The workaround of adding the function/variable to a library, fiddling with Package-Requires, and running `package-lint-current-buffer` repeatably until we got the answer is just to cumbersome.

purcell:
> The workaround of adding the function/variable to a library, fiddling with Package-Requires, and running `package-lint-current-buffer` repeatably until we got the answer is just to cumbersome.

I'm not quite clear on how this loop ends up happening, and I wonder whether a potential (or better) solution might be to include more info about symbols directly in the `package-lint` output.

tarsius:
Oooch, I was being dramatic. Linting once should be enough.

The command you added serves my needs perfectly and its a nice surprise it even provides history.

PR #182, opened 2020-08-13
Add package-lint-check-symbol feature

conao3:
Add `package-lint-check-symbol` feature.
This feature suggested at #181.

To use this feature, `M-x package-lint-check-symbol cl-lib`,
then you get `cl-lib(features) is added at Emacs-24.3` in echo area.

If you use =package-lint-check-symbol= in lisp program like
`(package-lint-check-symbol 'cl-lib)`, it should return `'((features added (24 3)))`.

### DIFF
--- a/package-lint-test.el
+++ b/package-lint-test.el
@@ -750,5 +750,26 @@ Alternatively, depend on (emacs \"24.3\") or greater, in which cl-lib is bundled
     '((6 0 error "\"org-foobaz-test\" doesn't start with package's prefix \"ox-foobar\"."))
     (package-lint-test--run "(defun org-foobaz-test ()) (provide 'ox-foobar)" nil nil nil "ox-foobar" nil nil "ox-foobar"))))
 
+(ert-deftest package-lint-test-check-symbol ()
+  (should
+   (equal
+    '((features added (24 3)))
+    (package-lint-check-symbol 'cl-lib)))
+  (should
+   (equal
+    '((features added (24 4)))
+    (package-lint-check-symbol 'subr-x)))
+  (should
+   (equal
+    '((features added (24 4)))
+    (package-lint-check-symbol 'subr-x)))
+  (should
+   (equal
+    '((functions added (24 1))
+      (functions removed (26 1))
+      (variables added (24 1))
+      (variables removed (26 1)))
+    (package-lint-check-symbol 'gnus-sync-unload-hook))))
+
 (provide 'package-lint-test)
 ;;; package-lint-test.el ends here
--- a/package-lint.el
+++ b/package-lint.el
@@ -148,7 +148,7 @@ published in ELPA for use by older Emacsen.")
 
     (defconst package-lint--check-symbol-hash-table
       (let ((tbl (make-hash-table))
-            (tbl* (make-hash-table)))
+            (tbl-sorted (make-hash-table)))
         (pcase-dolist (`(,ver . ,info) stdlib-changes)
           (pcase-dolist (`(,kind (removed . ,removed) (added . ,added)) info)
             (dolist (r removed)
@@ -159,9 +159,9 @@ published in ELPA for use by older Emacsen.")
                    (let ((res (sort (nreverse value)
                                     (lambda (a b)
                                       (string< (symbol-name (car a)) (symbol-name (car b)))))))
-                     (puthash key res tbl*)))
+                     (puthash key res tbl-sorted)))
                  tbl)
-        tbl*)
+        tbl-sorted)
       "A hash-table contains symbol and added/removed info.")
 
     (defconst package-lint--check-symbol-hash-table-keys
--- a/package-lint.el
+++ b/package-lint.el
@@ -1182,7 +1182,7 @@ VER is list represents version."
   (mapconcat
    (lambda (elm)
      (pcase-let ((`(,kind ,status ,ver) elm))
-       (format "%s(%s) is %s at Emacs-%s"
+       (format "%s(%s) was %s in Emacs %s"
                symbol kind status
                (mapconcat #'number-to-string ver "."))))
    info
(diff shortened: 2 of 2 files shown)


conao3:
CI failed? I cannot handle it...

---

@tarsius, please try it, and please comment.

akirak:
@conao3 It might be due to `#` character in the branch name. It's used for comments in bash, so it's generally better to avoid it.

package-lint.el
```diff
@@ -1148,9 +1173,36 @@ otherwise invalid read forms are ignored."
...
 

 ;;; Public interface
 
+;;;###autoload
+(defun package-lint-check-symbol (symbol)
+  "Return SYMBOL is added/removed on what Emacs Version using package-lint database."
+  (interactive
+   (list (completing-read "Symbol: " package-lint--check-symbol-hash-table-keys)))
+  (let* ((symbol* (if (stringp symbol) (intern symbol) symbol))
```
purcell, line 1199: Here, just rebind `symbol`, since the `*` suffix just makes it easy to use the wrong one.

package-lint.el
```diff
@@ -1148,9 +1173,36 @@ otherwise invalid read forms are ignored."
...
+         (info (gethash symbol* package-lint--check-symbol-hash-table))
+         (msg (package-lint--check-symbol-make-message symbol info)))
```
purcell, line 1201: See, here you use `symbol`, when probably you needed to use `symbol*`.

package-lint.el
```diff
@@ -146,6 +146,31 @@ published in ELPA for use by older Emacsen.")
...
+            (dolist (a added)
+              (push `(,kind added ,ver) (gethash a tbl)))))
+        (maphash (lambda (key value)
+                   (let ((res (sort (nreverse value)
+                                    (lambda (a b)
+                                      (string< (symbol-name (car a)) (symbol-name (car b)))))))
+                     (puthash key res tbl*)))
+                 tbl)
+        tbl*)
+      "A hash-table contains symbol and added/removed info.")
```
purcell, line 165: I think all of this should be computed on the fly, rather than redundantly stored in memory just in case the query function is used.

package-lint.el
```diff
@@ -146,6 +146,31 @@ published in ELPA for use by older Emacsen.")
...
+
+    (defconst package-lint--check-symbol-hash-table-keys
+      (let (keys)
+        (maphash (lambda (key _value) (push key keys))
```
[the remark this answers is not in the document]
conao3: Thanks, I haven't known that ever.

package-lint.el
```diff
@@ -146,6 +146,31 @@ published in ELPA for use by older Emacsen.")
               stdlib-changes)
       "An alist of function/macro names and when they were removed from Emacs.")
 
+    (defconst package-lint--check-symbol-hash-table
+      (let ((tbl (make-hash-table))
+            (tbl* (make-hash-table)))
```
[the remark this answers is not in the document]
conao3: OK, rename variable name.

package-lint.el
conao3: OK, compute hashtable when the funtion invoke.

purcell:
Thanks for your work on this. I got a bit frustrated with the fact that the existing code was clunky and making it difficult to add this feature, so I ended up re-writing it and then implementing this new interactive command in #184. It just adds a handful of lines to the code in total, and the existing code has been greatly simplified. Check it out and let me know what you think.

package-lint-test.el
```diff
@@ -750,5 +750,26 @@ Alternatively, depend on (emacs \"24.3\") or greater, in which cl-lib is bundled
...
+   (equal
+    '((features added (24 4)))
+    (package-lint-check-symbol 'subr-x)))
+  (should
+   (equal
+    '((functions added (24 1))
+      (functions removed (26 1))
+      (variables added (24 1))
+      (variables removed (26 1)))
+    (package-lint-check-symbol 'gnus-sync-unload-hook))))
```
conao3, line 772: Thanks for implement another way to archive this feature!

I have two.
Point one is related to  `gnus-sync-unload-hook`, for example. @purcell's `package-lint-describe-symbol-history` replies only function?
But this package database has variable info too.
```
History of gnus-sync-unload-hook:
* function-removed in Emacs 26.1
* function-added in Emacs 24.1
```
---
Point two is I think it is better that the message bottom is newer.
Typically, messages are viewed in order from the top, so the bottom should be the most up-to-date information.
```
History of gnus-sync-unload-hook:
* function-removed in Emacs 26.1
* function-added in Emacs 24.1
```
should be 
```
History of gnus-sync-unload-hook:
* function-added in Emacs 24.1
* function-removed in Emacs 26.1
```
I think.

conao3:
Thanks for implement another way to archive this feature!

I have two.
Point one is related to gnus-sync-unload-hook, for example. @purcell's package-lint-describe-symbol-history replies only function?
But this package database has variable info too.

```
History of gnus-sync-unload-hook:
* function-removed in Emacs 26.1
* function-added in Emacs 24.1
```
https://github.com/purcell/package-lint/pull/182/files/c7f0d6b715dd6ba384b92339337704bf6f4a0d07#diff-85e79a11bde03dc130740f9604558b0fR766-R772

---

Point two is I think it is better that the message bottom is newer.
Typically, messages are viewed in order from the top, so the bottom should be the most up-to-date information.
```
History of gnus-sync-unload-hook:
* function-removed in Emacs 26.1
* function-added in Emacs 24.1
```
should be
```
History of gnus-sync-unload-hook:
* function-added in Emacs 24.1
* function-removed in Emacs 26.1
```
I think.

[Note] Resend this comment as a simple comment instead of a review comment.

purcell:
Yes, I agree. Both are done now.

conao3:
Thanks!

PR #193, opened 2020-10-19
defgroup check

dickmao:
In foo.el, `(defcustom foo-custom :group foo)` without corresponding `(defgroup foo)` should flag an error (implicit here is an insistence that `defgroup foo` reside in foo.el).

In foo.el, `(defcustom foo-custom :group whatever)` is allowed (and the reach of static analysis prevents us from verifying `whatever` is defgroup'ed).


purcell:
> implicit here is an insistence that `defgroup foo` reside in foo.el

This isn't really something we can assume, unfortunately.

PR #194, opened 2020-10-22
Check alias symbol from define-obsolete-{function, variable}-alias

conao3:
Hi.

```
(defalias 'foobar 'string-equal)
```

warns "Aliases should start with the package's prefix \"package-lint\". (emacs-lisp-package)".

On the other hand,

```
(define-obsolete-function-alias 'foobar 'string-equal)
```

generates

```
(progn
  (defalias 'foobar 'string-equal nil)
  (make-obsolete 'foobar 'string-equal nil))
```

but no warnings.
I think this macro should define alias with prefix.  How about think?

### DIFF
--- a/package-lint-test.el
+++ b/package-lint-test.el
@@ -674,6 +674,14 @@ Alternatively, depend on (emacs \"24.3\") or greater, in which cl-lib is bundled
     (package-lint-test--run "(defvaralias 'foobar 'string-equal)")))
   (should
    (equal
+    '((6 0 error "Aliases should start with the package's prefix \"test\"."))
+    (package-lint-test--run "(define-obsolete-function-alias 'foobar 'string-equal)")))
+  (should
+   (equal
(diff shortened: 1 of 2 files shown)


conao3:
Off topic: `Update stdlib-changes / build (pull_request)` is faild.
But the step is needed for every PRs? Running this step only Emacs minor update is sufficient?

purcell:
> I think this macro should define alias with prefix. How about think?

Obsolete aliases are a mechanism authors can use for temporary backward compatibility as they start to add package name prefixes to function and variable names. So the current behaviour is intentional and desirable.

> Off topic: `Update stdlib-changes / build (pull_request)` is faild.

Yes, I'll look into that when I get chance.

purcell:
Thanks for your work creating this PR though!

conao3:
OK, thanks!

PR #197, opened 2020-12-07
`package-lint--check-objects-by-regexp`: conses

### LINKED ISSUE #196: Wrong type argument, `listp` if `defalias`, ... are mapped in an alist
To reproduce:
1. Open an Elisp file containing an alist with a mapping like `(defalias . 5)` (as can be found in `lispy`)
2. `M-x package-lint-current-buffer`
3. Observe `wrong-type-argument` `listp`.

The cause of the issue is as follows:
`package-lint--check-objects-by-regexp` finds s-expressions beginning with `defalias`, that are in fact conses. Many checker functions passed to it then take the `cadr`, which fails, since the `cdr` isn't a list.

I have fixed this locally and will make a PR shortly.

nbfalcon:
The former function now only calls FUNCTION if the sexp it found is not
a cons whose `cdr' is not a list.

Fixes #196.

### DIFF
--- a/package-lint.el
+++ b/package-lint.el
@@ -1147,7 +1147,8 @@ otherwise invalid read forms are ignored."
       (let ((obj (unless (package-lint--inside-comment-or-string-p)
                    (save-excursion
                      (ignore-errors (read (current-buffer)))))))
-        (when obj (funcall function obj))))))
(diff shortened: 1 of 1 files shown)


nbfalcon:
The CI failure doesn't seem to be my fault.

PR #200, opened 2021-01-10
Update docs to refer to non-obsolete functions

tomfitzhenry:
package-lint-setup-flymake was deleted in https://github.com/purcell/package-lint/commit/bed4e7f58f7984b720140bd4fee81a3e163f55c0


purcell:
Thanks

Issue #83, opened 2017-08-19
Check for conforming licences


purcell:
All elisp packages for versions of Emacs that support `package.el` _must_ be licenced under GPLv3+ or compatible licences.

We should report errors if this cannot easily be determined to be the case by looking at package headers. Some guidelines about the text people should include is [in this comment](https://github.com/melpa/melpa/commit/c3366117f3399660509f62849a88dc8be56a1fef#commitcomment-23721000).

Tagging @tarsius here.

alphapapa:
Really?  Why is this?  Not that I object, just curious.

tarsius:
> Really? Why is this? Not that I object, just curious.

For the Emacsmirror I wrote some code to detect the license a long time ago and a few weeks ago I published some statistics on `emacs-devel`. I thought people would be happy that most packages are released under some license authored by the FSF. However it turned out that Richard was instead deeply troubled that there are packages that are not released under a GPLv3 compatible license, including GPLv2 licensed packages. (I do not share his opinion that because Emacs is releases under the GPLv3, all "This file is not part of Emacs" Elisp has to be released under a compatible license.)

https://www.youtube.com/watch?v=TmQKihNpsHk#t=3m19s

As a result I had to quickly improve my extraction tools (there were ~600 packages whose license it could not detect initially) and contact maintainers to ask them to change the license so we could continue to distribute them while keeping Richard happy.

That was my main intention when working on the license detection code; match as many weird ways of specifying the license so that the situation would as quickly as possible not look worse than it actually is and to have mostly reliable lists of maintainers to contact. And it was a mess I wrote a long time ago to begin with.

So it is not quite ready to be used by `package-lint` yet. Eventually I would like to get it into acceptable shape and then it can be used here. You can find it here: https://github.com/emacscollective/elx/tree/license-wip. You can of course use it now already, but note that I will continue to make significant changes in a rather disorganized fashion, including to the function signature.

Fanael:
But why? Packages can be use incompatible licenses just fine, all it would prevent is redistributing them with Emacs itself (and that's irrelevant because Emacs is not free software anyway, what with only the FSF being allowed to contribute); people can download and use them without issue.

To me, it sounds like needless pandering to the FSF.

tarsius:
[...]

PR #209, opened 2021-04-26
Lint License and SPDX-License-Identifier headers

lassik:
Implements issue #83 

Here's a first cut of the linter. Rules very much subject to debate:

* I copied the recommended license list from the issue, but it's probably not quite broad enough.
* Do we have the authority to require a `SPDX-License-Identifier:` header? I'd say it's okay because it's just a warning.
* `License:` is a non-standard header (even in Emacs) whereas `SPDX-License-Identifier:` is used by Linux and FreeBSD among many other projects, so if some license header is to be standardized, the latter would make more sense.

This patch doesn't detect the presence or absence of the GPL boilerplate in the headers. How important is that?

### DIFF
--- a/package-lint.el
+++ b/package-lint.el
@@ -89,6 +89,37 @@ The path can be absolute or relative to that of the linted file.")
     (buffer-string)))
 
 
+;;; Utilities
+
+(defun package-lint--region-paragraphs (start end)
+  "Return the paragraphs between START and END as string lists.
+
+Runs of whitespace in each paragraph are normalized to one space each."
+  (save-match-data
+    (save-excursion
+      (save-restriction
+        (let* ((whitespace "[ \t\r\n]")
+               (whitespace* (concat whitespace "*"))
+               (whitespace+ (concat whitespace "+"))
+               (para-break (concat "\n" whitespace* "\n" whitespace*))
+               (paragraphs '()))
+          (narrow-to-region start end)
+          (goto-char (point-min))
+          (while (not (eobp))
+            (when (looking-at whitespace+)
+              (goto-char (match-end 0)))
+            (let* ((pstart (point))
+                   (pend (or (and (re-search-forward para-break nil t)
+                                  (match-beginning 0))
+                             (point-max)))
+                   (para (replace-regexp-in-string
+                          whitespace+ " "
+                          (string-trim
+                           (buffer-substring-no-properties pstart pend)))))
+              (push para paragraphs)))
+          (reverse paragraphs))))))
+
+
 ;;; Machinery
 
 (defvar package-lint--errors nil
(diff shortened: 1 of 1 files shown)


lassik:
cc @tarsius who is very experienced with elisp licensing

tarsius:
> This patch doesn't detect the presence or absence of the GPL boilerplate in the headers. How important is that?

Extremely important.

1. The FSF instructs users of the license to use these permission statements to make their choice known.
2. The SPDX folks recommend that the spdx id is added *in addition* to the permission statement:

> Standard license headers
>
> When a license defines a recommended notice to attach to files under that license (sometimes called a “standard header”), the SPDX project recommends that the standard header be included in the files, in addition to an SPDX ID.
>
> Additionally, when a file already contains a standard header or other license notice, the SPDX project recommends that those existing notices should not be removed. The SPDX ID is recommended to be used to supplement, not replace, existing notices in files.
>
> Like copyright notices, existing license texts and notices should be retained, not replaced ‐ especially a third party’s license notices.
>
> The SPDX project’s statement regarding standard license headers and SPDX short-form identifiers can be found at Appendix V of the SPDX specification, version 2.1.
```

https://spdx.dev/ids

tarsius:
I've also worked on *this* a bit over the weekend. The Emacsmirror/`elx.el` now uses SPDX IDs if possible.

> I copied the recommended license list from the issue, but it's probably not quite broad enough.

https://emacsmirror.net/stats/licenses.html#orgaf6e078 is another list (it also includes non-spdx identifiers and I just noticed that ZLIB has to be normalized to Zlib. There may be other issues like that.

purcell:
Nice, @lassik! Agree with @tarsius that we should ideally recognise the recommended GPL boilerplate typically inserted by `auto-insert`: if it's there, but there are no formal licence headers, then we _should_ still accept this. (This is largely for backwards compatibility.)

Beyond that,

> Do we have the authority to require a SPDX-License-Identifier: header? I'd say it's okay because it's just a warning.

Sure, either that or `Licen[cs]e:`. Again I'd ideally accept the latter for general backwards compatibility in common cases but suggest the `SPDX-` form when necessary. And parsing any more than a handful of non-SPDX strings from `License:` wouldn't be required. Just thinking that `License: GPLv3` is probably good enough, for example.

lassik:
OK, GPL and LGPL boilerplate detection now added. There are probably subtle bugs in it, but seems to work.

lassik:
The current code needs a `SPDX-License-Identifier: ` to figure out which version of GPL or LGPL is being used.

Things would be easier for us on all fronts if we could require that header to always be present. Is there any harm from adding it, apart from the fact that GNU doesn't require it? The number of people using weird licenses not in SPDX is probably tiny.

lassik:
The current code in this PR doesn't check for absence of `at your option any later version` when using `GPL-x.x-only`.

But the `-only` licenses shouldn't be recommended, as they are incompatible with Emacs versions released under a later GPL.

lassik:
@tarsius Thanks for your work on https://emacsmirror.net/stats/licenses.html. 

Isn't `GPL-3.0-only` going to be problematic if future GPL versions are made? There's surprisingly many packages using it: 183. How do you detect `GPL-x.x-only` vs `GPL-x.x-or-later`?

tarsius:
> Isn't `GPL-3.0-only` going to be problematic if future GPL versions are made? There's surprisingly many packages using it: 183. How do you detect `GPL-x.x-only` vs `GPL-x.x-or-later`?

Most of those are probably due to `elx` finding no information in addition to the license file. I suspect that most of the authors who only use a license file do so because adding the permission statement is "unnecessary, noisy and ugly". Many would probably not object to `GPL-3.0-or-later` if they were made aware of the issue. They might still have reservations about the permission statements; so telling them about the less noisy `SPDX-License-Identifier: GPL-3.0-or-later` might help a lot.

lassik:
Could we convince GNU to bless `SPDX-License-Identifier` as the official way to specify the license for elisp packages?

tarsius:
I don't know about GNU and the FSF, but maybe emacs-devel. The former two surely are already aware of SPDX and so far they have not embraced it. But maybe the stars would be aligned if you brought it up now. Such things usually take a few attempts. (I would focus on just emacs-devel though.)

PR #212, opened 2021-05-09
More accurate regexp for looks-like-a-package-p

wyuenho:
So when I turn on `flycheck-package`, and if I'm looking at my init.el that happens to have a section called `;; Version Control`, package-lint won't think I'm looking at a elisp package.


purcell:
Thanks, I ended up using lm-get-header-re instead, and added a test

PR #214, opened 2021-05-11
Add "ol-" entry to package-lint--allowed-prefix-mappings

tarsius:
It's like `ob-` and `ox-`, just slightly newer, I think.


purcell:
Cool, thanks

Issue #111, opened 2018-09-08
Support multi-file packages


vermiculus:
I know this is a question that's out there already (#110), but I couldn't find an open issue and I'd like to keep track of it (i.e., subscribe to this issue) so I can make corresponding updates to [emake](https://github.com/vermiculus/emake.el/) when appropriate. As of now, I think I'm going to instruct emake to only `package-lint` the root package file with all the headers.

vermiculus:
Related conversation from the above issue:

> I don't think I want to encourage that kind of duplication -- especially things like package.el-sensitive headers like `Package-Requires` and `Version`.

Duplicating headers between all package files seems to ask for that information to become out-of-sync. (I *know* that I would be unable to keep them in sync without some sort of semi-official tooling.)

DamienCassou:
I change the headers across all files using wgrep. I also don't bother specifying all of the package dependencies in every file: only the minimum Emacs version to make package-lint happy.

purcell:
A little update here: I've been working on this (slowly) recently.

My idea would be that if `package-lint` is being run on a file-backed elisp buffer, it would look for a separate "master" package file in the same directory and use that as the source of package dependencies (and package name) for the buffer. Some thoughts around this:

- To find the "master" file, we proceed from the current file name (e.g. `my-package-utils.el`) and look for a corresponding `-pkg.el` file, otherwise we remove a dash-separated portion of the file name and repeat, this time looking for either `my-package.el` or `my-package-pkg.el`, continuing until we find package headers, or run out of shorter-named files to examine (in which case we assume the last non-`pkg` `.el` file encountered is the "master" file).
- An important case is when directories contain things like `my-package-flycheck.el` and `my-package.el`, each of which have package headers. If a package header is found in the current file, then it should be assumed to be the master file. This means that @DamienCassou's policy of putting package headers in all files would cause problems for the proposed mechanism.
- We don't have support for `-pkg.el` files in `package-lint` etc.
- When encountering `-pkg.el` files, we can provide an "info" message that the file is unnecessary for MELPA.
- `flycheck-package` currently doesn't bother to check a file if it doesn't have its own package headers, which would need to change.
- Also regarding `flycheck`/`flymake`, there's likely a small overhead when checking involves reading other files.

Goal would be to be able to be able to run `package-lint`'s batch checker on a bunch of files (either together or individually) and get a list of package issues that take into account the apparent package structure.

[...]

PR #218, opened 2021-09-01
New feature: autoload: package-lint-package, lint multiple files

Boruch-Baum:
+ lints all files matching a name-space in a specific directory

+ lints all files in a directory tree

+ configurable subdirectory exclusion list, by regexp

  + package-lint-package-exclude-subdirectories

### DIFF
--- a/package-lint.el
+++ b/package-lint.el
@@ -7,7 +7,7 @@
 ;; URL: https://github.com/purcell/package-lint
 ;; Keywords: lisp
 ;; Version: 0
-;; Package-Requires: ((cl-lib "0.5") (emacs "24.1") (let-alist "1.0.6"))
+;; Package-Requires: ((emacs "25.1") (let-alist "1.0.6"))
 
 ;; This program is free software; you can redistribute it and/or modify
 ;; it under the terms of the GNU General Public License as published by
--- a/package-lint.el
+++ b/package-lint.el
@@ -1223,6 +1223,67 @@ Current buffer is used if none is specified."
       (view-mode 1))
     (display-buffer buf)))
 
+(defcustom package-lint-package-exclude-subdirectories
+  (list "^\\..*$")
+  "Subdirectories not to search for lint-able files.
+Each element is a regular expression."
+  :group 'package-lint
+  :type '(repeat (regexp)))
+
+
+;;;###autoload
+(defun package-lint-package (path)
+  "Display lint errors and warnings for all files of a package.
+Prompts for PATH, which if a directory means all *.el files in
+that tree, and if a file means all files in that tree with that
+file's base-name as prefix. Default is the path of the current
+buffer, if an elisp file. Symbolic links to sub-directories are
+not followed. See also variable
+`package-lint-package-exclude-subdirectories'."
+  (interactive "fEnter path to main file: ")
+  (let* ((excludes
+           (lambda (x)
+             (let ((array package-lint-package-exclude-subdirectories)
+                   found elem)
+               (while (and (setq elem (pop array))
+                           (not (setq found (string-match elem x)))))
+               (not found))))
+         (dir   (if (file-directory-p path)
+                  (file-name-as-directory path)
+                 (file-name-directory path)))
+         (regex (format "^%s.*\\.el$"
+                        (if (file-directory-p path)
+                          ""
+                         (concat (file-name-base path) "-"))))
+         (files (directory-files-recursively dir regex nil excludes))
+         (buf "*Package-Lint*")
+         errs)
+    (unless (file-directory-p path)
+      (push path files))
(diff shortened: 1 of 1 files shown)


conao3:
Nice work!

package-lint.el
```diff
@@ -7,7 +7,7 @@
 ;; URL: https://github.com/purcell/package-lint
 ;; Keywords: lisp
 ;; Version: 0
-;; Package-Requires: ((cl-lib "0.5") (emacs "24.1") (let-alist "1.0.6"))
+;; Package-Requires: ((emacs "25.1") (let-alist "1.0.6"))
```
conao3, line 10: Why are you change minimum required Emacs version?
If you really want to this change, please create another PR.

Boruch-Baum: I performed package-lint on my modifications to package-lint and it gave me warnings that several of the functions I was using were from later Emacs versions. My memory is that function `directory-files-recursively' was what required 25.1. I thought to keep in the same PR because otherwise package-lint would fail itself, which would be embarrassing and ironic!

akirak:
@Boruch-Baum Because there can be a multi-package repository, i.e. a repository that contains source files for multiple packages in the same directory, you'll need a MELPA-style recipe to properly determine the files of a package. 

Both riscy's [melpazoid](https://github.com/riscy/melpazoid) and conao3's [keg.el](https://github.com/conao3/keg.el) do this, so there are already multiple solutions that run package-lint on an entire package, even though all of them require some additional configuration.

[package-build](https://github.com/melpa/melpa/blob/master/package-build/package-build.el) provides a function to retrieve a list of files from a directory and a recipe, which you could use to implement the proper version of `package-lint-package`. However, this means you will have to add `package-build` to the dependencies of `package-lint`.

Boruch-Baum:
On 2021-09-01 07:35, Akira Komamura wrote:
> ***@***.*** Because there can be a multi-package repository, i.e. a
> repository that contains source files for multiple packages in the same
> directory, you'll need a MELPA-style recipe to properly determine the
> files of a package.

Fundamentally, I'm not sure that's so relevant. The people most likely
to lint a repository are going to be the developers themselves, who I'm
imagining will have mostly be interested in linting one file at a time
or everything. They may not want to be part of the MELPA eco-system and
so that type of recipe might not exist.

Even for a MELPA curator, I'm not sure that your claim is true. With
this PR, if you select a file you only lint that file's name-space, not
the entire directory/repository. That should cover just the individual
package's elements within the repository, no?

If it does turn out to be the case that occasionally there is a
conflict, how common would that be? What percentage of cases does this
PR need to fail for it not to be an improvement? In this regard, I'd
distinguish between false-positives and false-negatives: A
false-positive I would consider a small pain, but easy to ignore;
false-negatives I would consider a nuisances. However, if it does need
to be 'hypothetically perfect' instead of 'great in practice', then I
guess the PR will be dropped!

Right now without a real-life example to test, my talk is hypothetical
Do you have a real-life failure example in mind that you can offer?

> Both riscy's [2]melpazoid and conao3's [3]keg.el do this, so there are
> already multiple solutions that run package-lint on an entire package,

I didn't know that! The only reason I made the effort is because on
another thread someone indicated that the feature did not exist ...

I'm not familiar with either melpazoid or keg ... Taking a first look,
it doesn't seem fair to compare package-lint to either of them. They may
be great, but ... melpazoid isn't even a MELPA package ?! And keg looks
like some kind of wishful parallel universe, with a very small adoption
via MELPA ...

> even though all of them require some additional configuration.

Ah. As so often the case, the devil is in the details ...

>  [4]package-build provides a function to retrieve a list of files from
>  a directory and a recipe, which you could use to implement the proper
>  version of package-lint-package. However, this means you will have to
>  add package-build to the dependencies of package-lint.

I'm not the package owner, but my sensibility would be to "go lean" if
at all possible and avoid that. If @purcell doesn't like the PR, that's
fine; he didn't ask me to do the work, so the effort is on me. Maybe it
can be the basis for someone else to do a better job.

To sum up: Can you point me to a problematic example?

--
hkp://keys.gnupg.net
CA45 09B5 5351 7C11 A9D1  7286 0036 9E45 1595 8BC0

akirak:
See also #111 for the previous discussions on support for multiple files.

Boruch-Baum:
On 2021-09-01 19:54, Akira Komamura wrote:
>    See also [1]#111 for the previous discussions on support for multiple
>    files.

Thanks. I'm being side-tracked just now by something urgent, so I'm
going to be busy with that for a bit, but thanks for that and your
previous mail about key that I also haven't been able to study.

--
hkp://keys.gnupg.net
CA45 09B5 5351 7C11 A9D1  7286 0036 9E45 1595 8BC0

PR #219, opened 2021-11-04
Consider define-advice forms to be global definitions

bcc32:
define-advice will define symbols that start with the name of the
function being advised, so they should be exempt from the package
prefix check.


purcell:
Thanks!

PR #223, opened 2022-02-12
Linting the linter

minad:
Fix long docstrings for Emacs 28. There is also the advantage that shorter
docstrings are display more nicely by my Marginalia annotation package.

I just saw this warnings pop up when I updated the package. Thank for this package and your work on MELPA! Think it gives a great boost to package quality across the ecosystem.


purcell:
Great, thank you!

PR #232, opened 2022-06-06
Accept Package-Requires header without version

kyanagi:
According to the manual, an entry in Package-Requires header may lack a version string.
An entry without a version is considered to have version "0".
https://www.gnu.org/software/emacs/manual/html_node/elisp/Library-Headers.html

So we should accept it.

### DIFF
--- a/package-lint-test.el
+++ b/package-lint-test.el
@@ -255,10 +255,6 @@ headers and provide form."
 (ert-deftest package-lint-test-error-invalid-dependency ()
   (should
    (member
-    '(6 0 error "Expected (package-name \"version-num\"), but found invalid.")
-    (package-lint-test--run ";; Package-Requires: (invalid)")))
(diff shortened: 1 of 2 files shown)


purcell:
You might be interested in the background to this warning: this is not legal syntax in all the versions of Emacs that have (or are supported by) `package.el`. So this means if one uses the syntax, packages will be reported as malformed on certain Emacs/package.el combinations, rather than simply uninstallable. As such, it makes sense for `package-lint` to direct users to avoid the less-compatible syntax.

However, based on the CI results, it _looks_ like that must have only have affected Emacs 23.4 and the very old standalone version of `package.el` which supported it, so we could choose to disregard this issue, and I think it would be a good idea to do so.

So I'll go ahead and merge this — thanks for taking the time to submit it.

PR #233, opened 2022-08-15
look for installable packages from the correct places

wyuenho:
`package-archive-contents` is empty until a `package-refresh-contents` and friends call is made to fetch from the archives, it also does not contain any builtin packages. This PR make `package-lint--check-packages-installable` look for the list of package descriptors from the builtins list, installed list and the archive cache.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -14,9 +14,6 @@ jobs:
     strategy:
       matrix:
         emacs_version:
-          - 24.1
-          - 24.2
-          - 24.3
           - 24.4
           - 24.5
           - 25.1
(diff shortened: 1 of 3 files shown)


purcell:
Hey - basic idea looks good, thanks! I need to think about a couple of issues:
- Lint results should be the same in any Emacs, so need to be careful about issues like that reported under Emacs 26.1 with these changes: `package-lint.el:10:38: warning: Version dependency for let-alist appears too high: try 1.0.5`. Here there is indeed a 1.0.6 in GNU ELPA, so this must be due to cached or built-in package info.
- Also need to think about any downstream effects of increasing the lower Emacs version bound for this library, as this change will break a bunch of CI configs — that's probably legitimate, but needs some sort of communication, and perhaps config examples.

purcell:
Unsure why CI is failing here with the error
```
package-lint-test.el:844:0: error: "Types" doesn't start with package's prefix "package-lint-test".
```
Probabaly related to the `imenu` code in `package-lint` — perhaps a problem was this was accidentally masked by the previous CI config.

purcell:
A change that should have the same effect as this is included in `package-lint` now, so I'll go ahead and close this — thanks!

Issue #125, opened 2019-06-24
Allow ignoring of errors


dajva:
Thanks for a great package.
Have you considered something like a white list or ignore list to allow specific constructs that violates some rule in `package-lint`?

This would allow to check major part of a package and new additions while keep older violating constructs for backwards compatibility.

I've found #94 but that was closed since the author changed his package, it seems.

purcell:
Thanks for asking! I'm not keen on this approach, really. `package-lint` is made for package authors, who therefore have the ability to change their code freely. The rules in `package-lint` are not opinionated: they follow the elisp conventions, or detect compatibility issues, so it doesn't make sense to be able to disable them IMO.

dajva:
Ok, thanks for the clarification.

noctuid:
Not allowing disabling errors assumes that package-lint is perfect, and it isn't (and neither is any other linter I've ever used). It gives a lot of false positives. I'd love to be able to fail a build if linting fails, but I can't do that if there isn't some way to silence incorrect warnings. Please reconsider this if it is feasible to add something like per-line ignore comments.

Here is just one example of invalid warnings:
```
2541:1: error: "use-package-normalize/:general" doesn't start with package's prefix "general".
2541:1: error: `use-package-normalize/:general' contains a non-standard separator `/', use hyphens instead (see Elisp Coding Conventions).
```
It is impossible to both integrate correctly with use-package and not have these errors. Package lint complains if `/` or `:` are in a function name even if they aren't actually used as an alternative separator to `-`. In this case you could maybe argue against using `/`, but I think it gives some visual clarity, and there is nothing I can do about it in my package.

purcell:
> It is impossible to both integrate correctly with use-package and not have these errors.

I can see how that's an issue for conventions like `use-package-normalize-` etc. There are exceptions for some such common ones coded into `package-lint`, and I would consider adding this one.

> In this case you could maybe argue against using `/`, but I think it gives some visual clarity, and there is nothing I can do about it in my package.

The point of the linter is exactly to combat the "I think X is better than the convention" argument, but if `use-package` mandates it then I can see that it is problematic in this specific case.

noctuid:
[...]

PR #242, opened 2023-03-13
Add use-package keyword definition prefixes to sane prefixes

noctuid:
Use-package requires using these prefixes to define new keywords.

Following up on discussion in #125.


purcell:
Thanks!

PR #244, opened 2023-03-15
Better handle obsolete aliases

purcell:
WIP


purcell:
Explicit tests just to be safe.

Issue #144, opened 2019-10-20
Avoid false positives wrt. "bad" symbols


purcell:
Code like [this](https://github.com/akirak/nix-env-install/blob/8b7501d6ac3f0dbed4b4fe844616dd88e9df875e/nix-env-install.el#L73-L77) (or `let` lists) can cause false positives for `package-lint` checks (as [here](https://github.com/melpa/melpa/pull/6452#issuecomment-544189657)).

A solution would be to double-check that apparent function names aren't instead in certain positions in containing forms, but this could get fiddly/inefficient.

nbfalcon:
#196 is slightly related.

PR #246, opened 2023-05-24
Don't confuse let bindings with function calls

purcell:


purcell:
See #144, among possibly others.

Issue #227, opened 2022-04-09
Support the new compat library


minad:
The [compat package](https://git.sr.ht/~pkal/compat) by @phikal provides backports of new APIs such that they can be used in packages which which are compatible with older Emacs versions. The list of backported APIs is given in the [MANUAL](https://git.sr.ht/~pkal/compat/tree/master/item/MANUAL) but probably one would also need some automatic API extraction as is used for the Emacs APIs itself.

purcell:
Nice, yes, that would quite take some thought to handle well, I think...

minad:
Sure, there is no hurry. compat was just released a few days ago. I think @tarsius also plans to adopt the library in magit and transient. However at this point the library will automatically see widespread usage and I assume that more authors are going to adopt it in order to avoid adding their own compatibility code. Furthermore packages get earlier access to new API additions. It seems as if core library development has accelerated/improved recently.

phikal:
Steve Purcell ***@***.***> writes:

> Thanks. It's certainly a useful initiative. I guess this is just for
> core functions etc., and won't work around the whole thing of people
> `require`-ing features that were only added to Emacs in a later
> version?

It is possible (though slightly dangerous as it involves advising
`require'), though as Compat isn't complete the only feature that is
supported that way is subr-x for Emacs 24.3.

-- 
	Philip Kaludercic

purcell:
Does sound a bit scary, but I don't think you'd have much choice. You could make people use `compat-require` instead, potentially, with that being a no-op in Emacs 28.

phikal:
Steve Purcell ***@***.***> writes:

> Does sound a bit scary, but I don't think you'd have much choice. You
> could make people use `compat-require` instead, 

That is a good idea, I hadn't considered it as setting aside subr-x, no
new features were implemented in Compat.

>                                                 potentially, with that
> being a no-op in Emacs 28.

No-op, or just a regular `require'?

-- 
	Philip Kaludercic

purcell:
> No-op, or just a regular `require'?

I (confusingly) meant "no-op" more as a direct pass-through, so yeah, just a regular `require`.

minad:
@purcell 

> Thanks. It's certainly a useful initiative. I guess this is just for core functions etc., and won't work around the whole thing of people require-ing features that were only added to Emacs in a later version?
> Does sound a bit scary, but I don't think you'd have much choice. You could make people use compat-require instead, potentially, with that being a no-op in Emacs 28.

[...]

PR #254, opened 2023-11-03
Add support for functions provided by the compat package

purcell:
See #227

### DIFF
--- a/package-lint.el
+++ b/package-lint.el
@@ -7,7 +7,7 @@
 ;; URL: https://github.com/purcell/package-lint
 ;; Keywords: lisp
 ;; Version: 0.19
-;; Package-Requires: ((cl-lib "0.5") (emacs "24.1") (let-alist "1.0.6"))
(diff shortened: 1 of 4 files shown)


package-lint.el
```diff
@@ -172,6 +172,30 @@ symbol such as `variable-added'.")
...
+  (let (symbols functions)
+    (dolist (ver '(25 26 27 28 29))
+      (let ((el-path (locate-library (format "compat-%d.el" ver) t)))
+        (unless el-path
+          (error "compat package not installed"))
+        (with-temp-buffer
+          (insert-file-contents el-path)
+          (goto-char (point-min))
+          ;; TODO convert to rx
+          (while (search-forward-regexp "^(compat-\\(defun\\|\\defvar\\) +\\_<\\(.*?\\)\\_>" nil t)
```
minad, line 185: Also compat-defmacro should be added here.

purcell: Ah yes! I should find an example and add a test.

purcell: And I've ignored variables, which aren't actually handled in the existing code.

PR #262, opened 2024-02-13
Add tools/compat-dump.el which generates data/compat-symbols

minad:
Load package-lint--compat-symbols from data/compat-symbols. Drop Compat dependency.


package-lint.el
```diff
@@ -172,33 +172,17 @@ symbol such as `variable-added'.")
...
-                                            (group (+? any))
-                                            symbol-end)
-                                        nil t)
-            (pcase (match-string 1)
-              ("defvar" (push (intern (match-string 2)) symbols))
-              ((or "defun" "defmacro" "defalias") (push (intern (match-string 2)) functions)))))))
-    (cons symbols functions))
+(defconst package-lint--compat-symbols
+  (with-temp-buffer
+    (insert-file-contents (expand-file-name "data/compat-symbols"))
```
purcell, line 177: This won't always find the file, so you should probably extract or duplicate the code that loads the sym-dump data

purcell:
That was quick :grin

minad:
> That was quick 😁

Well, all I had to do is to copy your extraction code. ;)

package-lint.el
minad: Right. I extracted a function `package-lint--load-data` which is used to load both data/compat-symbols and data/stdlib-changes.

PR #272, opened 2024-05-27
Remove face eldoc-highlight-function-argument from 25.1/removed

precompute:
This is a face.
Doc: "Face used for the argument at point in a function's argument list.  Note that this face has no effect unless the `eldoc-documentation-function' handles it explicitly."

I checked Emacs 25, 25.1, 26 and Master
25:
https://github.com/emacs-mirror/emacs/blob/emacs-25/lisp/emacs-lisp/eldoc.el#L113
25.1:
https://github.com/emacs-mirror/emacs/blob/f0eb70d8935be90f7c03e187c12d9b60e7214cc6/lisp/emacs-lisp/eldoc.el#L113
26:
https://github.com/emacs-mirror/emacs/blob/emacs-26/lisp/emacs-lisp/eldoc.el#L113
Master:
https://github.com/emacs-mirror/emacs/blob/master/lisp/emacs-lisp/eldoc.el#L141

The face is present in all of them.  This face is integral to the eldoc ux, so a removal and re-introduction seems unlikely.


purcell:
Okay, this file is produced programmatically by [dumping the symbols actually present in each Emacs version](https://github.com/purcell/package-lint/blob/master/tools/sym-dump.el), so any ad-hoc changes like this will quickly get overwritten. If you think there's something wrong with the linked script or associated logic, let me know, otherwise we should probably assume that the results are correct _despite_ your research.

precompute:
Thank you!  Sorry, I guess this should've been an Issue and not a Pull Request.
Nothing in the code that would lead to an error of this kind jumps out at me, so I'm going to close this pull request.  I haven't ran `sym_dump_all` yet, though.  If I do find something, I'll open a separate Pull Request / Issue.

PR #277, opened 2024-09-19
Convert run-tests.sh to Makefile

purcell:
Alternative formulation of #276, using my existing template for elisp project makefiles


purcell:
Thanks for the notes, the background helps. Some initial thoughts:
* The `init-package.el` shouldn't change much, so while it's annoying to edit inside a Makefile, it has allowed me to use the same makefile almost unchanged between dozens of elisp projects, so for me it's an okay trade-off
* For the SELECTOR thing, wouldn't you just run the ERT tests interactively inside Emacs, and then use ERT's commands to run individual tests if necessary? Doing this from the command line is a more unusual flow.
* Given that, the combined target for all tests isn't so bad

I'll have a think. :)

vHugoObject:
Running tests from the command line means I don't have to reevaluate changes I made to tests  to or the source code every time I make a change. For me, this is a particularly annoying problem as I prefer test-driven development. But even if you don't use test driven development this can become an issue. See this section from the ert manual: "If you have been editing and rearranging tests, it is possible that ERT remembers an old test that you have since renamed or removed: renamings or removals of definitions in the source code leave around a stray definition under the old name in the running process (this is a common problem in Lisp)."

purcell:
That makes sense. I incorporated your SELECTOR trick, and split the `test` target into two parts. I prefer this compromise overall because it's a little more compact, plus I've cleaned up some things in the old run-tests script that were obsolete, like ERROR_ON_WARN. Thanks for your help!

PR #280, opened 2025-01-06
Add org-babel-header-args to allowed prefixes

lhernanz:
Similar to the other org-babel prefixes already present, this one is also used by several modules.


purcell:
Thanks!

PR #286, opened 2025-02-25
Fix eshell/. being split over two lines

mattiasdrp:
Fixes #285

Issue #285, opened 2025-02-24
Debugger entered--Lisp error: (invalid-read-syntax "expected )" 1213 22)


MatthewZMD:
Using the latest commit from master, evaluating this function:

```lisp
(defconst package-lint-symbol-info
  (let* ((stdlib-changes (package-lint--load-data "data/stdlib-changes"))
         (info (make-hash-table)))
    (pcase-dolist (`(,version . ,data) stdlib-changes)
      (pcase-dolist (`(,syms . ,action)
                     (let-alist data
                       (list (cons .variables.added 'variable-added)
                             (cons .variables.removed 'variable-removed)
                             (cons .features.added 'library-added)
                             (cons .features.removed 'library-removed)
                             (cons .functions.added 'function-added)
                             (cons .functions.removed 'function-removed))))
        (dolist (sym syms)
          (puthash sym (cons (cons action version) (gethash sym info)) info))))
    info)
  "A hash table from SYMBOL to a list of events in its history.
Each event is of the form (ACTION . EMACS-VER), where ACTION is a
symbol such as `variable-added'.")
```

```lisp
Debugger entered--Lisp error: (invalid-read-syntax "expected )" 1213 22)
  read(#<buffer  *temp*>)
  (progn (insert-file-contents (expand-file-name file (if load-file-name (file-name-directory load-file-name) default-directory))) (read (current-buffer)))
  (unwind-protect (progn (insert-file-contents (expand-file-name file (if load-file-name (file-name-directory load-file-name) default-directory))) (read (current-buffer))) (and (buffer-name temp-buffer) (kill-buffer temp-buffer)))
  (save-current-buffer (set-buffer temp-buffer) (unwind-protect (progn (insert-file-contents (expand-file-name file (if load-file-name (file-name-directory load-file-name) default-directory))) (read (current-buffer))) (and (buffer-name temp-buffer) (kill-buffer temp-buffer))))
  (let ((temp-buffer (generate-new-buffer " *temp*" t))) (save-current-buffer (set-buffer temp-buffer) (unwind-protect (progn (insert-file-contents (expand-file-name file (if load-file-name (file-name-directory load-file-name) default-directory))) (read (current-buffer))) (and (buffer-name temp-buffer) (kill-buffer temp-buffer)))))
  package-lint--load-data("data/stdlib-changes")
[...]

PR #287, opened 2025-02-28
Override pp-default-function when generating stdlib-changes

pkryger:
With this fix, the data in `stdlib-data` is in single line, just like it used
to be before changes in #284.

I've run GitHub Actions in my fork to generate the new `stdlib-data`, please see [here](https://github.com/pkryger/package-lint/actions/runs/13586268268).

Similar to what d97dfc420e406f03b074315e2f7fdcd828cb0561 did. Inspired by [@charmitro's comment](https://github.com/purcell/package-lint/pull/286#issuecomment-2690064571).

Fixes: #285
Closes: #286


purcell:
Excellent, thanks — will merge once CI has finished.

This also fixes https://github.com/purcell/emacs.d/issues/873

PR #292, opened 2025-03-10
Bump the emacs-head-version

bbatsov:
Emacs 30 was released last month, so now the head version is 31.

I've noticed this problem when I tried running package-lint on a package requiring Emacs 30.

### DIFF
--- a/package-lint-test.el
+++ b/package-lint-test.el
@@ -270,7 +270,7 @@ headers and provide form."
   (should
    (equal
     '((6 23 warning "This makes the package uninstallable in all released Emacs versions."))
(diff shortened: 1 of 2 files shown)


bbatsov:
@purcell ping :-)

purcell:
Thanks. That change failed the corresponding test, but with the test fixed, we're good to go. Thanks for your patience, I've been having some downtime.

rrudakov:
Looks like it wasn't included in 0.25 release: https://github.com/purcell/package-lint/blob/0.25/package-lint.el

The version there is still 30.

purcell:
Just released 0.26, which includes this chnage.

Issue #282, opened 2025-01-21
`Package-Requires` header can span multiple lines


tarsius:
In `magit.el` I use this, for example:

```elisp
;; Package-Requires: (
;;     (emacs "27.1")
;;     (compat "30.0.1.0")
;;     (dash "2.19.1")
;;     (magit-section "4.2.0")
;;     (seq "2.24")
;;     (transient "0.8.2")
;;     (with-editor "3.4.3"))
```

(The trick is to put the open parenthesis on the same line as the keyword.)

With `package-lint` that results in:

```
magit.el:21:0: error: Couldn't parse "Package-Requires" header: \
  End of file during parsing
```

purcell:
Does that also work in older Emacsen, e.g. back to 24?

purcell:
ie. is it valid syntax for `package.el`, `lm-maint` etc in those older versions?

tarsius:
This was added in e8c46c2b6f76cc055366041b6112d61dd5f2dcf4, see https://debbugs.gnu.org/cgi/bugreport.cgi?bug=36301. So the first Emacs release to support it is 27.1.

Recently (bb5399e3cd75450db6db9b3c5829f7bd87ca1308, upcoming 30.1) this was moved to a new `lm-package-requires` function.

If a package specifies that it needs at least 27 or later, then multi-line `Package-Requires`, should not result in a linting warning. If it uses a multi-line value but does not also specify that 27.1 or later is required, then that should result in a warning.

purcell:
> If a package specifies that it needs at least 27 or later, then multi-line `Package-Requires`, should not result in a linting warning. If it uses a multi-line value but does not also specify that 27.1 or later is required, then that should result in a warning.

I don't really agree — you can't guarantee that package headers won't be parsed by older Emacsen, and in those versions you want Emacs to say, "Emacs 27 is not available", not "Malformed header". This is why we have a lint warning (error, even?) about `Package-Requires` entries that don't have versions, e.g. `(seq)`.

tarsius:
It's not guaranteed that this won't happen, but it is fairly unlikely. Nowadays, Melpa (and [Non]GNU Elpa) generate a `*-pkg.el` file and distribute a tarball, even for packages that consist of nothing but a single library. `package.el` will then only look at that file and ignore the library headers.

Old `package-install-file` and `package-install-from-buffer` would fail though.

It's up to you of course, but I think this should downgraded to a warning.

[...]

PR #295, opened 2025-03-27
Complete support for multi-line Package-Requires header

tarsius:
I proposed adding support for multi-line `Package-Requires` headers in #282.  The implementation added in #283 was incomplete and resulted in two new compiler warnings.  This pr finishes the implementation and pacifies the compiler.

### DIFF
--- a/package-lint.el
+++ b/package-lint.el
@@ -422,8 +422,10 @@ Instead it should use `user-emacs-directory' or `locate-user-emacs-file'."
   "Check the contents of the \"Package-Requires\" header.
 Return a list of well-formed dependencies, same as
 `package-lint--check-well-formed-dependencies'."
-  (let ((deps (package-lint--goto-header "Package-Requires")))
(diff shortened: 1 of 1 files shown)


purcell:
Nice, thanks! We'd need to add specific multiline tests too, including for the case of running package-lint in a somewhat-older Emacs.

tarsius:
> We'd need to add specific multiline tests too

Sure, and if you had added such tests in #283, this bug likely would not have landed. However, I don't think that it makes sense to hold of on fixing that bug by merging this, until someone™ has written the tests. You can manually test the effects of this bugfix by comparing the outputs of `package-lint-current-buffer` in `ghub.el` -- applying this pr reduces the number of false-positives from 62 to just 1.

purcell:
Let's merge this now, and I'll write a test in due course. 😁

tarsius:
Thanks!

PR #302, opened 2025-09-24
Add `package-lint-prefix` for explicit prefix configuration

countvajhula:
### Problem

Currently, `package-lint` infers the required symbol prefix from the filename, either directly for single-file packages or via `package-lint-main-file` for traditional multi-file packages.

However, this mechanism doesn't support the "package suite" pattern employed by some large projects, where multiple, distinct packages (e.g., `my-project-core`, `my-project-ui`) are developed in a single repository and share a common project prefix (e.g., `my-project-`). Linting the modules in `my-project-ui` will incorrectly expect a `my-project-ui-` prefix.

### Solution

This PR introduces a new `defvar`, `package-lint-prefix`. When this variable is non-nil, its string value is used as the definitive prefix for symbol checks, overriding any inferred prefix.

This provides a simple, robust, and explicit way for downstream tooling (like MELPA's melpazoid tool) to configure the linter for this "package suite" structure.

### Implementation

The change is implemented in two commits:

The first commit refactors the test suite to decouple buffer setup from the test runner. This enabled me to add a new test for the existing `package-lint-main-file` functionality, which was previously untested.

The second commit introduces `package-lint-prefix`, updates the checking logic to prioritize it, and adds tests demonstrating its behavior, including its precedence over `package-lint-main-file`.

### Motivation & Use Cases

This pattern is followed explicitly by [Symex](https://github.com/drym-org/symex.el) in its 2.0 release, which refactors formerly monolithic functionality into 5 separate, composable packages. It is also used in practice by major packages like Helm and Magit. Supporting this structure in `package-lint` would pave the way for other large Emacs packages that could benefit from this modular, maintainable design.

### Example Usage

The intended use is for client tools to bind this variable. For example, a MELPA recipe could gain a new keyword (e.g., `:project` or `:domain`):

```
(symex-core :domain "symex" ...)
```

Tools like melpazoid could then read this keyword and invoke the linter with the correct configuration:

```
(let ((package-lint-prefix "symex"))
  (package-lint-buffer "symex-core.el"))
```

### DIFF
--- a/package-lint-test.el
+++ b/package-lint-test.el
@@ -970,6 +970,39 @@ Alternatively, depend on (emacs \"24.3\") or greater, in which cl-lib is bundled
         (package-lint-test--run "(defun secondary-file-fn ())"
                                 :featurename "secondary-file"))))))
 
+(ert-deftest package-lint-test-explicit-prefix ()
+  "Test the `package-lint-prefix' variable."
+  (let ((package-lint-prefix "my-suite"))
+    ;; A correct, explicitly-provided prefix passes,
+    ;; even when the filename suggests a different prefix.
+    (should
+     (equal
+      '()
+      (package-lint-test--run "(defun my-suite-component-a-fn ())"
+                              :featurename "component-a")))
+
+    ;; A symbol using the filename's prefix fails when
+    ;; `package-lint-prefix' is set to something else.
+    (should
+     (equal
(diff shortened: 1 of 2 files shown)


countvajhula:
Thank you, Steve. This is a very clear and sensible response, and I completely agree with your philosophy of keeping the tool simple and strict. I really appreciate you taking the time to engage thoughtfully on this. Cheers!

PR #310, opened 2026-08-07
Support relative package-lint-main-file for files in a sub dir in batch mode

jian-lin:
See the newly added ERT test `package-lint-test-accept-relative-main-file-in-batch-mode` for how to reproduce the bug.  Specifically, the sub test "lint files in a sub dir" fails without the fix in this PR and shows the bug.

### DIFF
--- a/package-lint-test.el
+++ b/package-lint-test.el
@@ -933,6 +933,19 @@ Alternatively, depend on (emacs \"24.3\") or greater, in which cl-lib is bundled
     "(let ((bar |foo)) blah)"
     (package-lint--is-a-let-binding))))
 
+(ert-deftest package-lint-test-accept-relative-main-file-in-batch-mode ()
+  (let ((this-dir (if load-file-name
+                      (file-name-directory load-file-name)
+                    default-directory)))
+    (ert-info ("lint files in the current dir")
+      (let ((default-directory
+             (expand-file-name "batch-tests/relative-package-lint-main-file/" this-dir)))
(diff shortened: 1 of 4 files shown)


package-lint-test.el
```diff
@@ -933,6 +933,19 @@ Alternatively, depend on (emacs \"24.3\") or greater, in which cl-lib is bundled
     "(let ((bar |foo)) blah)"
     (package-lint--is-a-let-binding))))
 
+(ert-deftest package-lint-test-accept-relative-main-file-in-batch-mode ()
+  (let ((this-dir (if load-file-name
+                      (file-name-directory load-file-name)
+                    default-directory)))
```
jian-lin, line 942: Maybe use `ert-resource-directory` from `ert-x.el` instead.

package-lint-test.el
```diff
@@ -933,6 +933,19 @@ Alternatively, depend on (emacs \"24.3\") or greater, in which cl-lib is bundled
...
+    (ert-info ("lint files in the current dir")
```
jian-lin, line 943: Hmm, interesting.  This sub test, which is not what this PR fixes, already fails on Emacs 25, according to CI https://github.com/purcell/package-lint/actions/runs/31161834607/job/93133036318?pr=310#step:4:140

jian-lin: I pushed a new commit to mark the test as expected failure on Emacs 25, to make CI happy.

jian-lin:
CI is green now. PTAL @purcell

PR #311, opened 2026-08-07
Autoload the safe-local-variable property of package-lint-main-file

jian-lin:
This makes the safe-local-variable property of package-lint-main-file take effect before package-lint is loaded, as recommended by (info "(elisp) File Local Variables").


purcell:
Great, thanks.

